PR #1, opened 2019-03-18
Add MI distribution building artifacts

a5anka:


distribution/src/assembly/filter.properties
```diff
@@ -0,0 +1,27 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+product.name=WSO2 Micro Integrator
+product.key=MI
+product.version=6.4.0
```
sdkottegoda: Should we change the version to the next milestone?

a5anka: :+1:

PR #7, opened 2019-03-19
 Add org.wso2.carbon.micro.integrator.server from product-ei

a5anka:
Fix related to #2


components/org.wso2.carbon.micro.integrator.server/pom.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  ~ Copyright (c) 2016, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
madushadhanushka: Old license headers

a5anka: I suppose having 2016 is ok given that we are moving module instead of creating new ones.

PR #54, opened 2019-04-09
Bootstrap test framework

a5anka:


integration/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>2.20.1</version>
```
Nirothipan: we have this version defined in parent pom also

a5anka: Fixed in https://github.com/wso2/micro-integrator/pull/54/commits/f91b8f6d58f48c9cb7ba6fd6cd6cc0f42b52b1f4

Issue #41, opened 2019-04-04
Reorganize the files in the distribution


a5anka:
Currently, the file structure in the distribution is very similar to the one we had with the micro integrator profile. Since we are creating a separate build for micro integrator we can move most of these files to top level. 

- [x] Document folder structure
  - [x] Update micro-integrator/doc/working-with-the-micro-integrator.md

a5anka:
Copying external jars to `<MI_HOME>/lib directory` does not work. This needs to be fixed as well.

malakaganga:
> Copying external jars to `<MI_HOME>/lib` directory does not work. This needs to be fixed as well.

I have fixed this issue with

https://github.com/wso2/micro-integrator/pull/68

and tested with a Handler and Class mediator
Also, I have verified the patch applying process as well and it is working fine.

a5anka:
@malakaganga @nandika There are pending tasks in the issue. Therefore reopening the issue.

a5anka:
@malakaganga We get an additional log in the startup after #68

```
[2019-04-16 10:21:46,615] [micro-integrator]  INFO - DefaultCryptoProviderComponent 'CryptoService.Secret' property has not been set. 'org.wso2.carbon.crypto.provider.SymmetricKeyInternalCryptoProvider' won't be registered as an internal crypto provider. Please set the secret if the provider needs to be registered.
```

Can you have a look?

a5anka:
@malakaganga wso2server.sh and micro-integrator.sh does the same thing. Therefore it makes more sense to merge the two. We can just remove the micro-integrator.sh and rename wso2server.sh to micro-integrator.sh since the existing micro-integrator.sh does not do anything useful.

malakaganga:
Fixed https://github.com/wso2/micro-integrator/issues/41#issuecomment-483612868
With https://github.com/wso2/micro-integrator/pull/92

Fixed https://github.com/wso2/micro-integrator/issues/41#issuecomment-483616786

> @malakaganga wso2server.sh and micro-integrator.sh does the same thing. Therefore it makes more sense to merge the two. We can just remove the micro-integrator.sh and rename wso2server.sh to micro-integrator.sh since the existing micro-integrator.sh does not do anything useful.

with 
https://github.com/wso2/micro-integrator/pull/90

Hence resolving the issue

a5anka:
I am reopening this since we have not documented the folder structure anywhere which is a listed subtask of the issue.

malakaganga:
[...]

PR #68, opened 2019-04-12
Fix wso2/micro-integrator/issues/41

malakaganga:
## Purpose
Fixes:  https://github.com/wso2/micro-integrator/issues/41

### DIFF
--- a/distribution/src/conf/wso2server.sh
+++ b/distribution/src/conf/wso2server.sh
@@ -218,14 +218,14 @@ if [ "$jdk_17" = "" ]; then
 fi
 
 CARBON_XBOOTCLASSPATH=""
-for f in "$CARBON_HOME"/../lib/xboot/*.jar
+for f in "$CARBON_HOME"/wso2/lib/xboot/*.jar
 do
     if [ "$f" != "$CARBON_HOME/wso2/lib/xboot/*.jar" ];then
         CARBON_XBOOTCLASSPATH="$CARBON_XBOOTCLASSPATH":$f
     fi
 done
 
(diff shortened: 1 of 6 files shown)


distribution/src/conf/wso2server.sh
```diff
@@ -237,10 +237,13 @@ do
...
     fi
 done
-for t in "$CARBON_HOME"/../lib/commons-lang*.jar
+for t in "$CARBON_HOME"/wso2/lib/commons-lang*.jar
 do
     CARBON_CLASSPATH="$CARBON_CLASSPATH":$t
 done
+
+
+
```
a5anka: Do we need all these extra lines?

p2-profile/pom.xml
```diff
@@ -81,6 +81,9 @@
                                 <featureArtifactDef>
                                     org.wso2.carbon.mediation:org.apache.synapse.wso2.feature:${carbon.mediation.version}
                                 </featureArtifactDef>
+                                <featureArtifactDef>
+                                    org.wso2.carbon.crypto:org.wso2.carbon.crypto.feature:1.0.3
```
a5anka: Let's not hard code the version here.

a5anka:
Let's try to use a meaningful title for PRs. The one used in this one is not readable for humans (maybe readable for machines :) ).

distribution/src/conf/wso2server.sh
malakaganga: Fixed with : https://github.com/wso2/micro-integrator/pull/90

PR #81, opened 2019-04-17
Minor fixes related to Docker and k8s artifacts

a5anka:


distribution/src/resources/dockerfiles/Dockerfile
```diff
@@ -20,5 +20,5 @@
 
 FROM wso2/micro-integrator:latest
 
-COPY carbonapps /home/wso2ei/wso2mi/wso2/micro-integrator/repository/deployment/server/carbonapps
+COPY files/carbonapps /home/wso2ei/wso2mi/wso2/micro-integrator/repository/deployment/server/carbonapps
```
sdkottegoda: the location to carbonapps is not correct

a5anka: Fixed

PR #94, opened 2019-04-18
Add argument for start synapse unit test framework agent

sajithaliyanage:
## Purpose
Passing parameters to start the unit testing framework for synapse agent

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/distribution/src/scripts/micro-integrator.sh
+++ b/distribution/src/scripts/micro-integrator.sh
@@ -136,8 +136,8 @@ do
           CMD="version"
     elif [ "$c" = "--restart" ] || [ "$c" = "-restart" ] || [ "$c" = "restart" ]; then
           CMD="restart"
-    elif [ "$c" = "--test" ] || [ "$c" = "-test" ] || [ "$c" = "test" ]; then
-          CMD="test"
+    elif [ "$c" = "--synapseTest" ] || [ "$c" = "-synapseTest" ] || [ "$c" = "synapseTest" ]; then
+          CMD="synapseTest"
     else
         args="$args $c"
     fi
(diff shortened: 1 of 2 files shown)


distribution/src/scripts/micro-integrator.sh
```diff
@@ -138,6 +138,8 @@ do
           CMD="restart"
     elif [ "$c" = "--test" ] || [ "$c" = "-test" ] || [ "$c" = "test" ]; then
           CMD="test"
+    elif [ "$c" = "--unitTest" ] || [ "$c" = "-unitTest" ] || [ "$c" = "unitTest" ]; then
```
sdkottegoda: what's the difference between the parmeter test and unitTest? What are the 2 modes?

sajithaliyanage: **test** tag is a previously added tag which I think use to execute JDK by file path. (under test tag **JAVACMD="exec "$JAVACMD""** command runs). **unitTest** is the tag which needs to start unit testing framework server parallel to the synapse engine using user given TCP port or default port. If a user wants to start the unit testing framework in product-ei, user can execute one of the following commands.
```
sh integrator.sh -unitTest
sh integrator.sh -unitTest <port>
```
If a user doesn't input any port, the unit testing server starts on the default port (9007).

sdkottegoda: So the 'test' argument doesn't actually test anything? If so, shall we remove the argument since it would be confusing to the users?

sajithaliyanage: Ok, I will check it and remove test tag if there no usage from it.

sajithaliyanage: Removed **test** parameter from the script file (micro-integrator.sh) and changed unitTest tag to **synapseTest**.

a5anka: @sajithaliyanage shall we document these somewhere to help the users?

a5anka: A guide on how to do unit testing will be ideal.

sajithaliyanage: @a5anka I will add a guide on how to do the unit test with MI. Shall I add it to the doc folder?

sajithaliyanage:
After upgrading the synapse to version 2.1.7-wso2v103 please merge this PR.

distribution/src/scripts/micro-integrator.sh
a5anka: :+1: 

Make sure a user can navigate to that page from the root readme. It does not have to be a direct link. :)

Issue #141, opened 2019-04-30
Prometheus support


a5anka:
We need to add Prometheus support to MI as well. We have done it for integrator profile in https://github.com/wso2/product-ei/pull/3399.

- [x] Doc updates
  - [x] Update [Getting started guide](https://github.com/wso2/micro-integrator/tree/master/examples/getting-started) with details (related to https://github.com/wso2/micro-integrator/issues/133)
  - [x] Update [Working with the Micro Integrator](https://github.com/wso2/micro-integrator/blob/master/doc/working-with-the-micro-integrator.md#working-with-the-micro-integrator)

PR #146, opened 2019-05-02
Add Prometheus support to Micro Integrator

Nirothipan:
Relevant Issue:- https://github.com/wso2/micro-integrator/issues/141

   https://github.com/wso2/product-ei/pull/3399


Nirothipan:
WIP - Do Not Merge

distribution/src/conf/synapse.properties
```diff
@@ -66,3 +66,7 @@ mediation.flow.statistics.collect.all=false
 
 # Script Mediator Pool (impatcts only external scripts)
 #synapse.script.mediator.pool.size=15
+
+# Internal APIs
+internal.http.api.enabled=true
```
a5anka: We have a strange character at the end of the line. Shall we remove it?

a5anka: Is there a way to hide this property to improve the user experience. For a user, this property does not make sense.

a5anka: Is there a way to hide this property to improve the user experience. For a user, this property does not make sense.

doc/working-with-the-micro-integrator.md
```diff
@@ -349,4 +349,75 @@ If you want to change the default locations of the registry directories, uncomme
...
+alerts. It collects statistical data exposed over an HTTP endpoint in the form of multi dimensional time series data,
+ which can then be visualized and queried. Since its inception, Prometheus has become a widely used monitoring tool among various companies and organizations, as it can be easily incorporated into existing systems, and is also integrable with popular graphical tools such as Grafana to visualize system behavior. Furthermore, Prometheus has its own querying language ‘PromQL’, and an expression browser to view basic graphical representations of data.
+
+The main objective of this feature is to facilitate users to make use of an extensively used system monitoring tool 
+to analyze the behavior of WSO2 Micro Integrator.  
+
+![prometheus_intro](images/prometheus_intro.png)
+
+
+#### Prometheus Publisher Component
```
a5anka: Will this section be useful for an enduser? Shall we remove this since this is some internal implementation details?

a5anka:
There is a typo in the new commit message. Shall we fix it?

distribution/src/conf/synapse.properties
Nirothipan: > We have a strange character at the end of the line. Shall we remove it?

This has happened since the original file had new line charecters ^M and git uses \n by default.
https://stackoverflow.com/questions/1889559/git-diff-to-ignore-m.

I have modified the original file without ^M in https://github.com/wso2/micro-integrator/pull/146/commits/9c983aa8bea14762ecde8b73ed4b60d08364362e and added my changes in https://github.com/wso2/micro-integrator/pull/146/commits/8ee685037c6624efd2c3cba960b79ec55eeb91e0. 

Originally I didn't modify the whole file since it is going to be a huge change, anyhow as we are going to encounter this again if we edit this file , I've modified the whole file.

Nirothipan:
> There is a typo in the new commit message. Shall we fix it?

Fixed

doc/working-with-the-micro-integrator.md
Nirothipan: removed in https://github.com/wso2/micro-integrator/pull/146/commits/34302a85e654ed95c64f04695cf2e99e767f2ff9

distribution/src/conf/synapse.properties
Nirothipan: > Is there a way to hide this property to improve the user experience. For a user, this property does not make sense and hard to imagine how this relates to Prometheus.

Since we have enabled this by default user don't need to worry about this property.
> 
> We also have another issue where the admin services also depend on the same property. Due to security reasons, we have to disable it by default. Therefore we cannot set this to true. One option is to comment the admin service related class in `internal-apis.xml`.

We will have to look into the possibility of this. @isudana suggestions please

> I am not sure whether the framework is flexible enough to start multiple internal APIs in multiple ports. 

Currently not AFAIK, however since we can differentiate based on url paths, it won't be an issue I guess. @isudana please correct if I'm wrong.


>When you look at it from the outside just having the system property to disable and enable Prometheus backend should be enough and the behaviour should not change on other configs. But here even when the system property is set if this property is false the backend will not work.

Yes true, even if the system property is set, BE will not work if this is disabled as per current implementation. We may need to change the implementation ( of both Internal API and Prometheus ) to enable internal API when this system property is set.

distribution/src/conf/internal-apis.xml
```diff
@@ -0,0 +1,23 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+<!--
+    This is the configuration file for internal APIs
+-->
+
+<apis>
+    <api name="MetricAPI" class="org.wso2.carbon.prometheus.publisher.service.MetricAPI"/>
```
isudana: Should it be Prometheus API?

distribution/src/conf/synapse.properties
isudana: > > Is there a way to hide this property to improve the user experience. For a user, this property does not make sense and hard to imagine how this relates to Prometheus.
> 
> Since we have enabled this by default user don't need to worry about this property.
> 
> > We also have another issue where the admin services also depend on the same property. Due to security reasons, we have to disable it by default. Therefore we cannot set this to true. One option is to comment the admin service related class in `internal-apis.xml`.
> 
> We will have to look into the possibility of this. @isudana appreciate your input on this please
> 
We cannot disable by default all internal APIs depends on this. Enabling internal API port is not related to Prometheus and it should be enabled by default. Enabling Prometheus has to be carried out with some other parameter. AFAIR there is a system parameter to do that. 
The problem here is, including internal API enabling stuff into this PR which might mislead @a5anka to think that it's related to Prometheus.

> > I am not sure whether the framework is flexible enough to start multiple internal APIs in multiple ports.
> 
> Currently not AFAIK, however since we can differentiate based on url paths, it won't be an issue I guess. please correct if I'm wrong.
> 
> > When you look at it from the outside just having the system property to disable and enable Prometheus backend should be enough and the behaviour should not change on other configs. But here even when the system property is set if this property is false the backend will not work.
> 
> Yes true, even if the system property is set, BE will not work if this is disabled as per current implementation. We may need to change the implementation ( of both Internal API and Prometheus ) to enable internal API when this system property is set.

distribution/src/conf/internal-apis.xml
Nirothipan: Yes , I'll rename this . anyway internally in mediation this is implemented as MetricAPI

distribution/src/conf/synapse.properties
Nirothipan: https://github.com/wso2/carbon-mediation/pull/1236 && https://github.com/wso2/carbon-mediation/pull/1235

PR #155, opened 2019-05-06
Adding analytics server configurations to carbon.xml

sajithshn:


distribution/src/conf/carbon.xml
```diff
@@ -649,5 +649,27 @@
 	    <RepositoryName>default repository</RepositoryName>
 	    <RepositoryURL>http://product-dist.wso2.com/p2/carbon/releases/wilkes/</RepositoryURL>
     </FeatureRepository>
+    
+       <!--
```
a5anka:
```suggestion
   <!-- 
```

Issue #175, opened 2019-05-08
Script mediator with registry reference throws exception when invoked


arunans23:
**Description:**

In micro-integrator, when a script mediator refers to a script in registry, it does not work and it throws an exception.

```
[2019-05-08 20:19:13,932] [micro-integrator] ERROR - ScriptMediator The script engine returned an Exception executing the external js script : Value {name ='null', keyValue ='gov:custom/checkJsScript.js'} function checkLog
java.lang.NullPointerException
	at org.wso2.carbon.mediation.registry.MicroIntegratorRegistry.lookUpFileMediaType(MicroIntegratorRegistry.java:983)
	at org.wso2.carbon.mediation.registry.MicroIntegratorRegistry.readNonXML(MicroIntegratorRegistry.java:816)
	at org.wso2.carbon.mediation.registry.MicroIntegratorRegistry.lookup(MicroIntegratorRegistry.java:203)
	at org.apache.synapse.registry.AbstractRegistry.getResource(AbstractRegistry.java:66)
	at org.apache.synapse.config.SynapseConfiguration.getEntry(SynapseConfiguration.java:867)
	at org.apache.synapse.core.axis2.Axis2MessageContext.getEntry(Axis2MessageContext.java:260)
	at org.apache.synapse.mediators.bsf.ScriptMediator.prepareExternalScript(ScriptMediator.java:486)
	at org.apache.synapse.mediators.bsf.ScriptMediator.mediateWithExternalScript(ScriptMediator.java:332)
	at org.apache.synapse.mediators.bsf.ScriptMediator.invokeScript(ScriptMediator.java:288)
	at org.apache.synapse.mediators.bsf.ScriptMediator.mediate(ScriptMediator.java:258)
	at org.apache.synapse.mediators.AbstractListMediator.mediate(AbstractListMediator.java:109)
	at org.apache.synapse.mediators.AbstractListMediator.mediate(AbstractListMediator.java:71)
	at org.apache.synapse.mediators.base.SequenceMediator.mediate(SequenceMediator.java:158)
	at org.apache.synapse.rest.Resource.process(Resource.java:363)
	at org.apache.synapse.rest.API.process(API.java:431)
	at org.apache.synapse.rest.RESTRequestHandler.apiProcess(RESTRequestHandler.java:135)
	at org.apache.synapse.rest.RESTRequestHandler.dispatchToAPI(RESTRequestHandler.java:113)
[...]

PR #177, opened 2019-05-09
Add file extensions to mediaType extensions in registry.xml

arunans23:
Related to https://github.com/wso2/micro-integrator/issues/175

Since micro-integrator registry is different from WSO2 registry, it does not have all the metadata related to the registry files. In Script mediator, it requires a [DataHandler](https://github.com/wso2/wso2-synapse/blob/master/modules/extensions/src/main/java/org/apache/synapse/mediators/bsf/ScriptMediator.java#L494) which in turn requires [media type](https://github.com/wso2/carbon-mediation/blob/master/components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/WSO2Registry.java#L704) of the uploaded registry file. As a temporary fix, we need to map all the file extensions with Content-Type so that we can retrieve the media-type in runtime by looking at the file extension.

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/components/org.wso2.carbon.micro.integrator.core/pom.xml
+++ b/components/org.wso2.carbon.micro.integrator.core/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>org.wso2.ei</groupId>
         <artifactId>mi-component-parent</artifactId>
-        <version>1.0.0-m7</version>
+        <version>1.0.0-m8-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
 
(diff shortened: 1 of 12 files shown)


distribution/src/conf/registry.xml
```diff
@@ -28,4 +28,24 @@
     <parameter name="GovRegRoot">{Root directory path for governance Registry}</parameter>
     <parameter name="LocalRegRoot">{Root directory path for local Registry}</parameter>
     -->
+
+    <!--
+        Since MicroIntegrator has a filed based registry, the following mapping is used to get the media-type of the file
+    -->
+    <parameter name="fileExtensionMappings"><![CDATA[
```
a5anka: It's not nice to have CDATA tag to specify xml elements. Shall we try to fix this?

arunans23: Moving file mappings to code

PR #217, opened 2019-05-17
cli: change name of compressed archive

menuka94:
Change the name of the compressed archive from `mi` to `wso2mi-cli`.

### DIFF
--- a/cmd/build.sh
+++ b/cmd/build.sh
@@ -125,7 +125,7 @@ do
 
     echo -en "\t - $goos/$goarch..."
 
-    zipfile="$filename-$build_version-$pos-$parch"
+    zipfile="wso2$filename-cli-$build_version-$pos-$parch"
     zipdir="${buildPath}/$filename"
     mkdir -p $zipdir
 


cmd/build.sh
```diff
@@ -125,7 +125,7 @@ do
 
     echo -en "\t - $goos/$goarch..."
 
-    zipfile="$filename-$build_version-$pos-$parch"
+    zipfile="wso2$filename-cli-$build_version-$pos-$parch"
```
Nirothipan: @menuka94  What does variable **filename** refers to here ? Can't we change that itself ?

This will change only the external zip name and after extracting we will have folder **mi** and **mi.exe** inside. Aren't we changing that too as **wso2mi-cli**.
@a5anka Thoughts please ?

menuka94: > @menuka94 What does variable **filename** refers to here ? Can't we change that itself ?

No. `$filename` refers to the main `.go` file (in this project it's `mi.go`. If we change that to `wso2mi-cli.go`, then the CLI itself becomes `wso2mi-cli <command>`.

> This will change only the external zip name and after extracting we will have folder **mi** and **mi.exe** inside. Aren't we changing that too as **wso2mi-cli**.

That could be done if needed.

a5anka: @Nirothipan the purpose was to just change the archive name to avoid confusion.

Nirothipan:
@menuka94  The names are still as mi-* not **wso2mi-cli** Jenkins build job. Could you please check https://wso2.org/jenkins/job/products/job/micro-integrator/106/

menuka94:
> @menuka94 The names are still as mi-* not **wso2mi-cli** Jenkins build job. Could you please check https://wso2.org/jenkins/job/products/job/micro-integrator/106/

That is because the change has been overridden and reverted to the original state by a8ca53cf030059cb55150b623346646649e90c74. See the new PR #244.

PR #226, opened 2019-05-21
Add org.wso2.carbon.rule.mediation.server.feature

tharinduwijewardane:
Pack the rule mediation feature with MI. 
Add config file rule-engine-config.xml.

Fixes wso2/product-ei#4041

Dependent PR: https://github.com/wso2/carbon-rules/pull/39

### DIFF
--- a/distribution/src/assembly/bin.xml
+++ b/distribution/src/assembly/bin.xml
@@ -386,6 +386,13 @@
             <fileMode>644</fileMode>
         </file>
 
+        <file>
+            <source>
+                ../p2-profile/target/wso2carbon-core-${carbon.kernel.version}/wso2/conf/rule-engine-config.xml
+            </source>
+            <outputDirectory>wso2mi-${pom.version}/conf</outputDirectory>
+            <fileMode>644</fileMode>
+        </file>
 
         <file>
             <source>src/conf/axis2.xml</source>
--- a/p2-profile/pom.xml
+++ b/p2-profile/pom.xml
@@ -442,6 +442,11 @@
                                     <version>${carbon.commons.version}</version>
                                 </feature>
 
+                                <!--10 Carbon rule-->
+                                <feature>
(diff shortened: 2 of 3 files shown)


distribution/src/conf/rule-engine-config.xml
```diff
@@ -0,0 +1,12 @@
+<ruleEngineConfig xmlns="http://wso2.org/carbon/rules/conf">
```
Nirothipan: Shall we add the licence header for this file

distribution/src/conf/rule-engine-config.xml
```diff
@@ -0,0 +1,12 @@
...
+    <!--<ruleEngineProvider className="org.wso2.carbon.rule.backend.jsr94.JSR94BackendRuntimeFactory">
+        <property name="propertyProvider"
+                  value="org.wso2.carbon.rule.backend.jsr94.drools.DroolsJSR94DefaultPropertiesProvider"/>
```
Nirothipan: @tharinduwijewardane  Is there any reason to specifically adding this file here rather than suing the one coming from the kernel.
https://github.com/wso2/product-ei/blob/master/distribution/src/assembly/bin.xml#L2333.

distribution/src/conf/rule-engine-config.xml
Nirothipan: Is there any specific reason for adding this file here. It will be available with the feature itself https://github.com/wso2/carbon-rules/blob/master/features/mediation/org.wso2.carbon.rule.mediation.server.feature/resources/conf/rule-engine-config.xml. Can't we just pack that.

tharinduwijewardane: Modified the bin.xml file to pack the config file which is available with the feature.
Thanks for pointing out.

tharinduwijewardane:
This increases the file size of MI by 7.8 MB

PR #234, opened 2019-05-27
cli: match version from build environment

### LINKED ISSUE #121: Use the MI version as the CLI version
Currently, the version used for MI CLI is different from the product version. Shall we use the same version for CLI too since it is easier to correlate?

We need to make sure the CLI version upgrade also happens whenever we do an MI release.

menuka94:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,11 +14,32 @@
 # limitations under the License.
 # -----------------------------------------------------------------------
 
+# do not indent (Makefile syntax, not bash)
+ifdef MVN_RELEASE_VERSION
+# maven release
+$(info MVN_RELEASE_VERSION is $(MVN_RELEASE_VERSION))
+VERSION=$(MVN_RELEASE_VERSION)
+else
+# not a maven release
+ifdef POM_VERSION
+VERSION=$(POM_VERSION)
+$(info POM_VERSION is $(POM_VERSION))
+endif
+endif
+# one of these will be set to VERSION in Jenkins Production Environment
+
+
 .PHONY: install
 install:
 	mvn clean install
-	cd cmd && ./build.sh -t mi.go -v 1.0.0 -f;
+	cd cmd && ./build.sh -t mi.go -v ${VERSION} -f
 
 .PHONY: install-cli
 install-cli:
-	cd cmd && ./build.sh -t mi.go -v 1.0.0 -f
+	cd cmd && ./build.sh -t mi.go -v ${VERSION} -f
+
+.PHONY: install-cli-local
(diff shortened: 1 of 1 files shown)


Nirothipan:
@menuka94  this is intend to fix https://github.com/wso2/micro-integrator/issues/121 , isn't it

menuka94:
> @menuka94 this is intend to fix #121 , isn't it

Yes. I'll update the commit message.

a5anka:
As discussed offline, shall we try to read the version from pom.file for local builds?

menuka94:
> As discussed offline, shall we try to read the version from pom.file for local builds?

Done.

Makefile
```diff
@@ -14,11 +14,32 @@
...
+ifdef POM_VERSION
+VERSION=$(POM_VERSION)
+endif
+endif
+# one of these will be set to VERSION in Jenkins Production Environment
+
+
+ifndef VERSION
+# local build, use the POM version
+VERSION:=$(shell mvn -q -Dexec.executable=echo -Dexec.args='$${project.version}' --non-recursive exec:exec)
```
a5anka: Why do we have two `$` characters?

a5anka:
@menuka94 Do we need to read from the Jenkins environment now?

Makefile
menuka94: Because the expression inside parentheses needs to get evaluated and assigned to variable `VERSION`.

menuka94:
> @menuka94 Do we need to read from the Jenkins environment now?

Yes, because otherwise the version will not match the Maven release version.

a5anka:
> > @menuka94 Do we need to read from the Jenkins environment now?
> 
> Yes, because otherwise, the version will not match the Maven release version.

Then shall we have two make targets? One to use in Jenkins and one to use when building locally.

menuka94:
> > > @menuka94 Do we need to read from the Jenkins environment now?
> > 
> > Yes, because otherwise, the version will not match the Maven release version.
> 
> Then shall we have two make targets? One to use in Jenkins and one to use when building locally.

Done.

a5anka:
@menuka94 Can you let me know when we can merge this without breaking the product build?

menuka94:
> @menuka94 Can you let me know when we can merge this without breaking the product build?

Sure. Even if this is merged before deploying the new Maven Release Plugin in Jenkins, general builds will work fine. The only caveat is that in release builds, the CLI will not have the new release version.

menuka94:
@a5anka This can be merged now.

PR #242, opened 2019-05-30
Add registry migration tool

tharinduwijewardane:
Please find more information in the attached README.md file.

Dependent PR: https://github.com/wso2/carbon-mediation/pull/1270

### DIFF
--- a/migration/registry/from-mi-1.0.0-to-later/README.md
+++ b/migration/registry/from-mi-1.0.0-to-later/README.md
@@ -1,0 +1,37 @@
+#Registry Migration Tool for WSO2 Micro Integrator
+
+This tool creates metadata files for registry resources. 
+Metadata files which are used to store media type of the resource were not available in the MI 1.0.0 release.
+The tool infers media types from the file extensions of the resource files.
+
+##How to run the tool
+
+1. Build RegistryMigrator using maven 
+
+`mvn clean package`
+
+2. Run the jar file providing the path to MI distribution 
+
+`java -jar target/RegistryMigrator-1.0.jar <Absolute_Path_To_MI_HOME>`
+
+##More information
+
+The metadata related to registry resource is stored in a separate file named <resource_file_name>.meta inside 
+the ".metadata" dir which is located where the registry resource is.
+At the moment media type is the only metadata stored.
+
+Example registry structure:
+
+    |registry/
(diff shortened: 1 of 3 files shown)


tharinduwijewardane:
The default behavior for files whose extension is not in the map is to create the metadata file anyway and add the media type as `text/plain`. 
@a5anka, Is that ok or should we skip those files?

a5anka:
> The default behavior for files whose extension is not in the map is to create the metadata file anyway and add the media type as `text/plain`.
> @a5anka, Is that ok or should we skip those files?
I'm OK with both options.

Shall we add a pom file to generate a runnable binary?

tharinduwijewardane:
> Shall we add a pom file to generate a runnable binary?

Done

migration/registry/from-mi-1.0.0-to-later/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+##How to run the tool
+
+1. Build RegistryMigrator using maven 
+
+`mvn clean package`
+
+2. Run the jar file providing the path to MI distribution 
+
+`java -jar target/RegistryMigrator-1.0.jar <Absolute_Path_To_MI_HOME>`
```
a5anka: Shall we point to the registry location instead of the product location?

This is due to the usage of shared file systems for the registry.

migration/registry/from-mi-1.0.0-to-later/README.md
```diff
@@ -0,0 +1,37 @@
+#Registry Migration Tool for WSO2 Micro Integrator
```
Nirothipan: Hi @tharinduwijewardane 

Seems the Mark Downs are not applied correctly ( https://github.com/wso2/micro-integrator/tree/master/migration/registry/from-mi-1.0.0-to-later ) . Shall we fix the readme file to look better.
https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#headers

a5anka: +1. We need to add a space after `#` in headings.

tharinduwijewardane: Fixed in https://github.com/wso2/micro-integrator/pull/246. Please approve.

PR #253, opened 2019-06-05
cli: Implement listing data-services using the management API

### LINKED ISSUE #220: Improve CLI to list DSS artifacts.

menuka94:


cmd/cmd/show.go
```diff
@@ -55,7 +58,17 @@ var showCmdHelp = `
...
-var showCmdValidArgs = []string{"api", "carbonapp", "endpoint", "inboundendpoint", "proxyservice", "sequence", "task", "help"}
+var showCmdValidArgs = []string{
+	"api",
+	"carbonapp",
+	"endpoint",
+	"inboundendpoint",
+	"proxyservice",
+	"sequence",
+	"task",
+	"data-service",
```
a5anka: This is different from `proxyservice`. Can we make it consistent?

cmd/cmd/root.go
```diff
@@ -29,7 +29,7 @@ import (
 var cfgFile string
 var verbose bool
 
-var programName = os.Args[0]
+var programName = "mi"
```
a5anka: We shouldn't change this. This is done to make the cli name dynamic. 
@SanojPunchihewa

cmd/cmd/show.go
menuka94: What shall we use? I'm in favor of using a hyphen.

a5anka: Since dataservice is a term specific to integration, It is ok to use without the hyphen.

cmd/cmd/dataServiceInfo.go
```diff
@@ -0,0 +1,110 @@
...
+	"fmt"
+	"github.com/spf13/cobra"
+	"github.com/wso2/micro-integrator/cmd/utils"
+)
+
+var dataServiceNameInput string
+
+const showDataServiceCmdLiteral = "dataservice"
+const showDataServiceCmdShortDesc = ""
+const showDataServiceCmdLongDesc = ""
```
a5anka: Shall we add a short and long description?

PR #267, opened 2019-06-13
Update distribution folder structure of the CLI

### LINKED ISSUE #266: MI binary should be inside the bin folder
Currently, the distribution tree is like below

```
.
├── LICENSE
└── mi
```

Shall we make it like below since most other programs adapt to this structure?

```
.
├── bin
│   └── mi
└── LICENSE
```

SanojPunchihewa:
## Purpose
- Move the executable to a bin directory
- Append the version to the extracted directory

Fixes #266

### DIFF
--- a/cmd/build.sh
+++ b/cmd/build.sh
@@ -153,8 +153,8 @@ do
 
     echo -en "\t - $goos/$goarch..."
 
-    mi_archive_name="wso2$filename-cli-$build_version-$pos-$parch"
     mi_dir_name="wso2$filename-cli-$build_version"
+    mi_archive_name="$mi_dir_name-$pos-$parch"
     mi_archive_dir="${buildPath}/${mi_dir_name}"
     mkdir -p $mi_archive_dir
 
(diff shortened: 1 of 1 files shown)


cmd/build.sh
```diff
@@ -154,14 +154,17 @@ do
     echo -en "\t - $goos/$goarch..."
 
     mi_archive_name="wso2$filename-cli-$build_version-$pos-$parch"
-    mi_archive_dir="${buildPath}/$filename"
+    mi_dir_name="wso2$filename-cli-$build_version"
```
a5anka: Shall we have this as a variable since it is used in multiple places?

cmd/build.sh
```diff
@@ -154,14 +154,17 @@ do
     echo -en "\t - $goos/$goarch..."
 
     mi_archive_name="wso2$filename-cli-$build_version-$pos-$parch"
```
a5anka: Shall we use `mi_archive_name="$mi_dir_name-$pos-$parch" instead?

SanojPunchihewa: Fixed in cd9c85e

PR #269, opened 2019-06-18
Restructure CLI Commands

SanojPunchihewa:
## Purpose
Change the ordering of the CLI commands to have a better categorization. Previously the format was,
`mi [verb] [noun] [parameters]` for example `mi show api [api-name]`
With this PR the format is changed to,
`mi [noun] [verb] [parameters]` for example `mi api show [api-name]`


a5anka:
@SanojPunchihewa Did we update the CLI documentation to match these changes?

cmd/cmd/api.go
```diff
@@ -22,16 +22,18 @@ import (
...
-	Use:   "completion",
-	Short: "Generates bash completion scripts",
-	Long:  `This command will generate a bash completion script`,
-	Run: func(cmd *cobra.Command, args []string) {
-		rootCmd.GenBashCompletionFile("mi_bash_completion.sh")
-	},
+// API command related usage info
+const apiCmdLiteral = "api"
+const apiCmdShortDesc = "Manage deployed Apis"
+const apiCmdLongDesc = "Manage the Apis deployed in the Micro Integrator"
```
a5anka: Do we need constants here? Are we using these anywhere else?

PR #280, opened 2019-06-19
Fix integration test cases for payload factory

GDLMadushanka:
Remove duplicated test cases


Nirothipan:
@GDLMadushanka why these files are deleted PayloadFactoryIntegrationSpecialCharactersAtPayloadFactoryTestCase.java

PR #337, opened 2019-06-21
fix test: NullNameSpaceForAttachPathTestCase

sajinieKavindya:
fix NullNameSpaceForAttachPathTestCase failure as part of the test hackathon


Nirothipan:
shall we copy these modules in pom

PR #353, opened 2019-06-24
Fix MediationStatEnableTestCase

GDLMadushanka:


integration/mediation-tests/tests-transport/src/test/resources/artifacts/ESB/server/repository/deployment/server/synapse-configs/default/proxy-services/rest-service-proxy.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
milindaperera: File name and the proxy name is different. Should be same to match standards

GDLMadushanka: +1 done

PR #358, opened 2019-06-24
Enabling Addressing module in the micro-integrator

rosensilva:


pom.xml
```diff
@@ -229,6 +229,12 @@
                 <artifactId>javax.jms-api</artifactId>
                 <version>${javax.jms-api.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.apache.axis2</groupId>
+                <artifactId>addressing</artifactId>
+                <type>mar</type>
+                <version>${addressing.version}</version>
```
Nirothipan: Shall we define this as axis2.wso2.version directly, since it is coming from there.

rosensilva: done

PR #365, opened 2019-06-24
Exclude admin tests related to message processor.

sachithKay:


Nirothipan:
@sachithKay lets remove the code itself since we are going to rewrite it for Management Apis. There will be no point in having them in the tests.

a5anka:
@sachithKay WDYT?

PR #378, opened 2019-06-24
Disable invalid test: IterateSequentialPropertySavingTestCase

sajinieKavindya:
Disable test runs for IterateSequentialPropertySavingTestCase since it checks the update scenario of a deployed proxy which is not valid in MI.


a5anka:
@rosensilva Shall we create a GitHub issue to track this?

PR #408, opened 2019-06-25
exclude test:DynamicLoadBalanceEpTestCase

sajithaliyanage:


chanikag:
Let's add a comment too to describe why these testcases are excluded.

sajithaliyanage:
Deleted the DynamicLoadBalanceEpTestCase.java file from the module due to no usage from it.

PR #417, opened 2019-06-25
Exclude spring mediator tests

sachithKay:


Nirothipan:
It would be better if we remove the classes and artifacts too. Else when remove the feature from pack, we may get errors if those artifacts are there in the server. WDYT .

milindaperera:
+1, let's remove testcases rather than commenting in testng.xml.
We already removed some tests of deprecated features

PR #456, opened 2019-06-26
Refactor applyConfiguration method

arunans23:
## Purpose
Refactor applyConfiguration method
    
Have a seperate method for applying configuration without restart


integration/tests-common/integration-test-utils/src/main/java/org/wso2/esb/integration/common/utils/common/ServerConfigurationManager.java
```diff
@@ -292,21 +292,17 @@ public void applyConfiguration(File newConfig) throws AutomationUtilException, I
...
-     * apply configuration file and optionally restart micro integrator server to take effect the configuration
+     * apply configuration file and restart micro integrator server to take effect the configuration
      *
      * @param newConfig       configuration file
-     * @param restartRequired optional Restart flag
      * @throws AutomationUtilException - throws if apply configuration fails
      * @throws IOException             - throws if apply configuration fails
      */
-    public void applyMIConfiguration(File newConfig, boolean restartRequired)
+    public void applyMIConfiguration(File newConfig)
```
Nirothipan: Shall we refactor the method names.
like appRlyMIConfiguration -> applyMIConfigurationWithRestart
applyConfigurationWithoutRestart -> appRlyMIConfiguration

Then we can easily in the PR itself when someone doing restart. Since restart is costly, this way we can prevent un intentional restarts.

arunans23: Fixed

Issue #366, opened 2019-06-24
Re-write test cases for admin services


sachithKay:
**Description:**
Test cases for admin services needs to be re-written after merging the micro-integrator management api.
Please refer: https://github.com/wso2/micro-integrator/pull/365

sajithaliyanage:
Test case needs to re-write - https://github.com/wso2/micro-integrator/pull/408

milindaperera:
Refer removal of local-entry related tests : https://github.com/wso2/micro-integrator/pull/425
API to retrieve deployed local-entries not available : https://github.com/wso2/micro-integrator/issues/426

sajinieKavindya:
Refer removal of TemplateAdminService related tests : [#438](https://github.com/wso2/micro-integrator/pull/438)

sajinieKavindya:
Refer removal of TemplateAdminService test: #438
Refer removal of RestApiAdminServiceTestCase and SequenceAdminServiceTestCase: #447

chanikag:
Exclude
org.wso2.carbon.esb.car.deployment.test.ESBJAVA3611EndpointTestCase
org.wso2.carbon.esb.car.deployment.test.ESBJAVA4116CAppArtifactDeploymentFailureDueToNameTestCase
org.wso2.carbon.esb.car.deployment.test.CarbonApplicationReDeploymentTestCase 
Related PR https://github.com/wso2/micro-integrator/pull/461

sachithKay:
Exclude Sample650TestCase. testLocalEntries.
https://github.com/wso2/micro-integrator/pull/473/

prabushi:
Exclude RedirectTest as it requires the Admin API to check the MS/MP status.
https://github.com/wso2/micro-integrator/pull/482/commits/d33303372bdf4c6bcc4eb731598291ca2938ac8f

prabushi:
API is required to check the whether a proxy service is faulty or not.
https://github.com/wso2/micro-integrator/pull/493/files

chanikag:
org.wso2.carbon.esb.resource.test.sequence.SequenceMediaTypeTestCase.dynamicSequenceMediaTypeTest
org.wso2.carbon.esb.resource.test.endpoint.EndPointMediaTypeTestCase.dynamicEndpointMediaTypeTest
org.wso2.carbon.esb.resource.test.template.EndpointTemplateMediaTypeTestCase.dynamicEndpointTemplateMediaTypeTest
[...]

PR #471, opened 2019-06-27
Fix capp deployment testcase failures

chanikag:
## Purpose
This PR is to fix capp deployment testcase failures


integration/mediation-tests/tests-patches/src/test/resources/testng.xml
```diff
@@ -148,10 +148,12 @@
...
         <classes>
-            <class name="org.wso2.carbon.esb.car.deployment.test.ESBJAVA3611EndpointTestCase"/>
+            <!-- Disable testcases as checking the service deployment is not supported
+            https://github.com/wso2/micro-integrator/issues/366 -->
+            <!--<class name="org.wso2.carbon.esb.car.deployment.test.ESBJAVA3611EndpointTestCase"/>-->
             <class name="org.wso2.carbon.esb.car.deployment.test.ESBJAVA3546CAppProxyOnLoadTestCase"/>
-            <class name="org.wso2.carbon.esb.car.deployment.test.ESBJAVA4116CAppArtifactDeploymentFailureDueToNameTestCase"/>
-            <class name="org.wso2.carbon.esb.car.deployment.test.CarbonApplicationReDeploymentTestCase"/>
+            <!--<class name="org.wso2.carbon.esb.car.deployment.test.ESBJAVA4116CAppArtifactDeploymentFailureDueToNameTestCase"/>-->
+            <!--<class name="org.wso2.carbon.esb.car.deployment.test.CarbonApplicationReDeploymentTestCase"/>-->
```
milindaperera: Shall we clean up this rather than commenting 
and let's delete the classes as well

integration/mediation-tests/tests-patches/src/test/java/org/wso2/carbon/esb/car/deployment/test/CARBON16113DeployArtifactsBeforeTransportStarsTestCase.java
```diff
@@ -42,45 +36,28 @@
...
     @BeforeClass(alwaysRun = true)
     private void initialize() throws Exception {
         super.init();
-        serverConfigurationManager = new ServerConfigurationManager(
-                new AutomationContext("ESB", TestUserMode.SUPER_TENANT_ADMIN));
+        carbonLogReader = new CarbonLogReader();
+        serverConfigurationManager = new ServerConfigurationManager(new AutomationContext());
+        carbonLogReader.start();
+        serverConfigurationManager.restartMicroIntegrator();
+        carbonLogReader.stop();
```
milindaperera: Are we starting and stoping log reader to assert startup log or missed to remove?

milindaperera: we can do this in test case rather than in init

integration/mediation-tests/tests-patches/src/test/resources/testng.xml
chanikag: This is to exclude the testcase as this functionality is not there at the moment. Once we resolve https://github.com/wso2/micro-integrator/issues/366 we can enable this and test this testcase. WDYT?

integration/mediation-tests/tests-patches/src/test/java/org/wso2/carbon/esb/car/deployment/test/CARBON16113DeployArtifactsBeforeTransportStarsTestCase.java
chanikag: In this testcase we get all the logs related to server startup, and therefore it is kind of a lengthy log string. if we do it later then it can be a large string and it takes more memory to evaluate. That's why in here, stop the log reader right after all the logs are retrieved.

Issue #474, opened 2019-06-27
[Admin API] API to retrieve deployed artifacts in registry


sajinieKavindya:
**Description:**
API to retrieve deployed artifacts in the registry is not available in current admin API

isudana:
This is available in the management API now.

PR #480, opened 2019-06-27
exclude test: DynamicURIWSDLEpTestCase

sajinieKavindya:
## Purpose
Exclude DynamicURIWSDLEpTestCase since there is no admin API to retrieve deployed artifacts in registry to test this case #474


chanikag:
Let's add this to https://github.com/wso2/micro-integrator/issues/366

PR #503, opened 2019-06-27
Exclude template test package

RidmiR:
## Purpose
Exclude template test package from the testng file. All these tests verify whether the resources are deploying correctly. There are admin APIs already implemented to check is those artifacts are exist and those are verifying under test services.

### DIFF
--- a/integration/mediation-tests/tests-other/src/test/resources/testng.xml
+++ b/integration/mediation-tests/tests-other/src/test/resources/testng.xml
@@ -91,14 +91,6 @@
         </packages>
     </test>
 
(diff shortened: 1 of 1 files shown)


milindaperera:
Let's remove class files as well.
Also, the Admin API does not contain the ability to retrieve deployed templates, hence let's create issue to include it

Nirothipan:
> Let's remove class files as well.
> Also, the Admin API does not contain the ability to retrieve deployed templates, hence let's create issue to include it

"Admin API does not contain the ability to retrieve deployed templates" I think this is already done in master. @sachithKay could you please confirm

PR #508, opened 2019-06-28
Fix ServerStartupTestCase testcase failure

chanikag:
## Purpose
This PR is to fix ServerStartupTestCase testcase failure and exclude testcases which needs management console utilities

### DIFF
--- a/integration/mediation-tests/tests-other/src/test/java/org/wso2/carbon/esb/ServerStartupTestCase.java
+++ b/integration/mediation-tests/tests-other/src/test/java/org/wso2/carbon/esb/ServerStartupTestCase.java
@@ -17,8 +17,30 @@
  */
 package org.wso2.carbon.esb;
 
-import org.wso2.carbon.integration.common.tests.ServerStartupBaseTest;
+import org.testng.Assert;
+import org.testng.annotations.BeforeSuite;
+import org.testng.annotations.Test;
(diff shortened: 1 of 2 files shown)


integration/mediation-tests/tests-other/src/test/java/org/wso2/carbon/esb/ServerStartupTestCase.java
```diff
@@ -17,8 +17,31 @@
...
+    }
+
+    @Test(groups = { "wso2.esb" }, description = "verify server startup errors")
+    @SetEnvironment(executionEnvironments = { ExecutionEnvironment.STANDALONE })
+    public void testVerifyLogs() {
+        carbonLogReader.start();
+        String logs = carbonLogReader.getLogs();
+        boolean status = logs.indexOf("Mgt Console URL") < logs.indexOf("Starting WSO2 Carbon");
+        carbonLogReader.stop();
+        Assert.assertFalse(status, "Server started with errors.");
```
a5anka: Shouldn't we search for exception stack traces here? Having `Mgt Console URL` and `Starting WSO2 Carbon` will not be enough to say that the server started without errors. Don't you think?

chanikag: Done in #575

PR #516, opened 2019-06-28
[Test Hackathon] Fix Class mediator test cases

arunans23:
## Purpose
Fix Class mediator test cases as part of Test Hackathon.


Nirothipan:
@arunans23  and @GDLMadushanka 
Here we have packed the synapse_samples jar directly in libs. This is packed in the product and the changes won't get reflected in test if we place it here. I think it would be better if we find the reason why it is not working when it is in plugins and fix it. Otherwise there will be no point of running the test case for that sample mediator.

arunans23:
> @arunans23 and @GDLMadushanka
> Here we have packed the synapse_samples jar directly in libs. This is packed in the product and the changes won't get reflected in test if we place it here. I think it would be better if we find the reason why it is not working when it is in plugins and fix it. Otherwise there will be no point of running the test case for that sample mediator.

The assumption here is that the code inside synapse_sample jar will not change. But if it changes, this approach won't work. If the purpose of this testcase is to check the functionality of the testcase, this fix would suffice.

Nirothipan:
> > @arunans23 and @GDLMadushanka
> > Here we have packed the synapse_samples jar directly in libs. This is packed in the product and the changes won't get reflected in test if we place it here. I think it would be better if we find the reason why it is not working when it is in plugins and fix it. Otherwise there will be no point of running the test case for that sample mediator.
> 
> The assumption here is that the code inside synapse_sample jar will not change. But if it changes, this approach won't work. If the purpose of this testcase is to check the functionality of the testcase, this fix would suffice.

If the assumption is the code inside samples won't change, why do we have a test case for it

arunans23:
> > > @arunans23 and @GDLMadushanka
> > > Here we have packed the synapse_samples jar directly in libs. This is packed in the product and the changes won't get reflected in test if we place it here. I think it would be better if we find the reason why it is not working when it is in plugins and fix it. Otherwise there will be no point of running the test case for that sample mediator.
> > 
> > 
> > The assumption here is that the code inside synapse_sample jar will not change. But if it changes, this approach won't work. If the purpose of this testcase is to check the functionality of the testcase, this fix would suffice.
> 
> If the assumption is the code inside samples won't change, why do we have a test case for it

My point is that the test case is to test the functionality of the Class mediator, not the sample jar.

Nirothipan:
> > > > @arunans23 and @GDLMadushanka
> > > > Here we have packed the synapse_samples jar directly in libs. This is packed in the product and the changes won't get reflected in test if we place it here. I think it would be better if we find the reason why it is not working when it is in plugins and fix it. Otherwise there will be no point of running the test case for that sample mediator.
> > > 
> > > 
> > > The assumption here is that the code inside synapse_sample jar will not change. But if it changes, this approach won't work. If the purpose of this testcase is to check the functionality of the testcase, this fix would suffice.
> > 
> > 
> > If the assumption is the code inside samples won't change, why do we have a test case for it
> 
> My point is that the test case is to test the functionality of the Class mediator, not the sample jar.

There are more jars like simple class mediator / class mediator demo etc. I think they also doing the same and this particular case is to test the test the samples.

arunans23:
> > > > > @arunans23 and @GDLMadushanka
> > > > > Here we have packed the synapse_samples jar directly in libs. This is packed in the product and the changes won't get reflected in test if we place it here. I think it would be better if we find the reason why it is not working when it is in plugins and fix it. Otherwise there will be no point of running the test case for that sample mediator.
> > > > 
> > > > 
> > > > The assumption here is that the code inside synapse_sample jar will not change. But if it changes, this approach won't work. If the purpose of this testcase is to check the functionality of the testcase, this fix would suffice.
> > > 
> > > 
> > > If the assumption is the code inside samples won't change, why do we have a test case for it
> > 
> > 
> > My point is that the test case is to test the functionality of the Class mediator, not the sample jar.
> 
> There are more jars like simple class mediator / class mediator demo etc. I think they also doing the same and this particular case is to test the test the samples.

But the name of the testcase and the name of the test method does not indicate that.

a5anka:
Packaging jar files directly in the repo is not maintainable and error-prone. IMO we should either,
- take external jars from a maven repository
- build internal jars from source and include it from a build step

PR #532, opened 2019-07-01
Check management api when starting server

sachithKay:


a5anka:
There is a DSS test failure in the Travis build. Shall we find the reason for failure?

sachithKay:
> There is a DSS test failure in the Travis build. Shall we find the reason for failure?

Fixed

Issue #536, opened 2019-07-01
Registry cache is not invalidated for properties unless the resource is updated


chanikag:
**Description:**
Steps to reproduce:
1. Set the cachableDuration to 0 in registry.xml
2. Add a proxy service to read a registry property and log the value
```
<proxy xmlns="http://ws.apache.org/ns/synapse"
       name="CachableDurationTestCaseProxy"
       transports="https http"
       startOnLoad="true"
       trace="disable">
    <description/>
    <target>
        <inSequence>
            <in>
                <property name="test" expression="get-property('registry','conf:/cache@cache_test')" scope="default"/>
                <log level="custom">
                    <property name="FROM" value="in"/>
                    <property name="test" expression="$ctx:test"/>
                </log>
                <drop/>
            </in>
            <out>
                <drop/>
            </out>
        </inSequence>
    </target>
</proxy>
```

3. In registry create files cache and cache.properties inside config registry
4. In   cache.properties file add an entry as 
cache_test=123456789

5. Start MI server.
6. Invoke the proxy service and following will be printed in log file.
INFO - LogMediator FROM = in, test = 123456789
7. While the server is running change the cache_test value in cache.properties file.
ex: cache_test=4666777868
8. Invoke the proxy service still the same value is printed.
9. Touch the cache file in registry
10. Then the new value is getting printed when the proxy service is invoked.
INFO - LogMediator FROM = in, test = 4666777868
**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Related Issues:**

PR #537, opened 2019-07-01
Fix CachableDurationTestCase testcase failure

chanikag:
## Purpose
This PR is to fix CachableDurationTestCase testcase failure.
The resource has to be updated to invalidate cache due to https://github.com/wso2/micro-integrator/issues/536


integration/tests-common/integration-test-utils/src/main/java/org/wso2/esb/integration/common/utils/MicroRegistryManager.java
```diff
@@ -100,6 +119,106 @@ public void updateResource(String resourcePath, String sourcePath, boolean backu
...
+        pathBuilder.append(System.getProperty("carbon.home")).append(File.separator).append("registry").append(File.separator);
+        if (collectionPath.startsWith(CONF_PATH)) {
+            String relativePath = collectionPath.substring(5).replace('/', File.separatorChar);
+            pathBuilder.append("config").append(relativePath);
+        } else if (collectionPath.startsWith(GOV_PATH)) {
+            String relativePath = collectionPath.substring(4).replace('/', File.separatorChar);
+            pathBuilder.append("governance").append(relativePath);
+        } else {
+            throw new MicroRegistryManagerException("Unknown resource path: " + collectionPath);
+        }
```
milindaperera: We can create a small function to above in else block to avoid code duplication

PR #539, opened 2019-07-01
test: exclude MediationStatEnableTestCase

menuka94:


GDLMadushanka:
Shall we explain the reason to exclude the test

menuka94:
> Shall we explain the reason to exclude the test

Updated the PR description.

PR #578, opened 2019-07-04
Fix intermittent test failures

chanikag:
## Purpose
This PR is to fix intermittent failures

### DIFF
--- a/integration/mediation-tests/tests-other/src/test/java/org/wso2/carbon/esb/statistics/PrometheusStatisticsTest.java
+++ b/integration/mediation-tests/tests-other/src/test/java/org/wso2/carbon/esb/statistics/PrometheusStatisticsTest.java
@@ -19,6 +19,7 @@
 package org.wso2.carbon.esb.statistics;
 
 import org.apache.http.HttpResponse;
+import org.awaitility.Awaitility;
 import org.testng.Assert;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
@@ -28,6 +29,7 @@ import org.wso2.esb.integration.common.utils.ESBIntegrationTest;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 public class PrometheusStatisticsTest extends ESBIntegrationTest {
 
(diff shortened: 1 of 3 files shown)


integration/mediation-tests/tests-other/src/test/java/org/wso2/carbon/esb/statistics/PrometheusStatisticsTest.java
```diff
@@ -40,6 +42,8 @@ public void init() throws Exception {
     @Test(groups = { "wso2.esb" }, description = "Test if metric data are exposed, when it is enabled in config")
     public void testPrometheusPublisher() throws Exception {
 
+        Awaitility.await().pollInterval(50, TimeUnit.MILLISECONDS).atMost(60, TimeUnit.SECONDS).
+                until(isManagementApiAvailable());
```
Nirothipan: Prometheus has a different API than Management. I added a fix for it in https://github.com/wso2/micro-integrator/pull/577

chanikag: It is the internal API right?
In the logs this is what you get
INFO  [org.wso2.carbon.automation.extensions.servers.utils.ServerLogReader] - [2019-07-03 18:20:25,091] [micro-integrator]  INFO - PassThroughListeningIOReactorManager Pass-through EI_INTERNAL_HTTP_INBOUND_ENDPOINT Listener started on 0.0.0.0:9391
INFO  [org.wso2.carbon.automation.extensions.servers.utils.ServerLogReader] - [2019-07-03 18:20:25,128] [micro-integrator]  INFO - PassThroughListeningIOReactorManager Pass-through EI_INTERNAL_HTTPS_INBOUND_ENDPOINT Listener started on 0.0.0.0:9354

If the HTTPS port there, the HTTP port is also available

Nirothipan: Yes , because we start http and then https . In this way also it will work . I added a check to check http port , as that is what we really want in this check .

PR #587, opened 2019-07-08
Add synapse xml configs to api and proxy response in Management API

sachithKay:


sachithKay:
Resolves: https://github.com/wso2/micro-integrator/issues/602

Issue #545, opened 2019-07-01
[Admin API] API resource to deactivate proxy service


RidmiR:
**Description:**
Need admin API service to disable proxy service in order to test the MI integration test of CAppDeactivateAndRestartTestCase

**Related Issues:**
https://github.com/wso2/micro-integrator/issues/366

PR #618, opened 2019-07-30
Refactor proxyservice resource to stop and start proxyservices

sachithKay:
Fix: https://github.com/wso2/micro-integrator/issues/545

### DIFF
--- a/components/org.wso2.carbon.micro.integrator.extensions/org.wso2.carbon.micro.integrator.management.apis/src/main/java/org/wso2/carbon/micro/integrator/management/apis/ProxyServiceResource.java
+++ b/components/org.wso2.carbon.micro.integrator.extensions/org.wso2.carbon.micro.integrator.management.apis/src/main/java/org/wso2/carbon/micro/integrator/management/apis/ProxyServiceResource.java
@@ -19,10 +19,14 @@
 
 package org.wso2.carbon.micro.integrator.management.apis;
 
+import com.google.gson.JsonObject;
 import org.apache.axiom.om.OMElement;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.synapse.MessageContext;
+import org.apache.synapse.ServerConfigurationInformation;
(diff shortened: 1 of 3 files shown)


components/org.wso2.carbon.micro.integrator.extensions/org.wso2.carbon.micro.integrator.management.apis/src/main/java/org/wso2/carbon/micro/integrator/management/apis/ProxyServiceResource.java
```diff
@@ -51,26 +61,42 @@ public ProxyServiceResource(String urlTemplate){
...
-            populateProxyServiceData(messageContext, param);
+        if (messageContext.isDoingGET()) {
+            String param = Utils.getQueryParameter(messageContext, "proxyServiceName");
+            if (Objects.nonNull(param)) {
+                populateProxyServiceData(messageContext, param);
+            } else {
+                populateProxyServiceList(messageContext);
+            }
+            axis2MessageContext.removeProperty(Constants.NO_ENTITY_BODY);
         } else {
```
sdkottegoda: I suppose the else condition is assuming a POST request? What if we get a DELETE/PUT request? Do we not need to respond with a valid message saying these operations are not valid.

Nirothipan: IIRC the not supported method calls will not propagate to here, it will be responded back in mediation ( carbon-mediation) level itself.

sachithKay: As nirothipan has mentioned, this resource will not receive any requests other than GET and POST.

PR #677, opened 2019-08-29
Fix class not def found issue

Nirothipan:

### DIFF
--- a/components/org.wso2.carbon.micro.integrator.initializer/src/main/java/org/wso2/carbon/micro/integrator/initializer/ServiceBusInitializer.java
+++ b/components/org.wso2.carbon.micro.integrator.initializer/src/main/java/org/wso2/carbon/micro/integrator/initializer/ServiceBusInitializer.java
@@ -50,8 +50,8 @@ import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.component.annotations.ReferenceCardinality;
 import org.osgi.service.component.annotations.ReferencePolicy;
-import org.wso2.carbon.base.ServerConfiguration;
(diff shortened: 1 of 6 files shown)


components/org.wso2.carbon.micro.integrator.initializer/src/main/java/org/wso2/carbon/micro/integrator/initializer/ServiceBusInitializer.java
```diff
@@ -217,7 +218,7 @@ protected void deactivate(ComponentContext ctxt) {
 
     private void initPersistence(SynapseConfigurationService synCfgSvc, String configName) throws AxisFault {
         // Initialize the mediation persistence manager if required
-        ServerConfiguration serverConf = ServerConfiguration.getInstance();
+        CarbonServerConfigurationService serverConf = CarbonServerConfigurationService.getInstance();
```
isudana: Can we not use Carbon word here?

isudana: and is it an osgi service?

Nirothipan: Used as we are reading configs from carbon.xml

Nirothipan: > is it an osgi service?

yes

sdkottegoda:
Closing since the fixes are duplicated in https://github.com/wso2/micro-integrator/pull/678

PR #694, opened 2019-09-02
Move  Transport Handlers to MI

Nirothipan:
This PR moves the  Transport Handlers to MI.
Renamed the pkg to  org.wso2.micro.integrator.transport.handlers.
Removed the carbon core service references.

This fixes the CNF when an API is invoked.


components/org.wso2.micro.integrator.transport.handlers/pom.xml
```diff
@@ -0,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+# Copyright (c) 2015, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
sdkottegoda: Incorrect date?

Nirothipan: updated

PR #722, opened 2019-09-06
move config mapper logic from product-ei

madushadhanushka:
Add single toml file configuration support for the MI.


sdkottegoda:
Now that https://github.com/wso2/micro-integrator/pull/774 is there, can we close this one?

PR #725, opened 2019-09-08
Refactor and arrange osgi wiring

milindaperera:

### DIFF
--- a/components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/internal/Activator.java
+++ b/components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/internal/Activator.java
@@ -95,6 +95,7 @@ public class Activator implements BundleActivator {
         registration = bundleContext.registerService(CarbonServerConfigurationService.class.getName(),
                                                      carbonServerConfiguration,
                                                      null);
+        CarbonCoreDataHolder.getInstance().setServerConfigurationService(carbonServerConfiguration);
         CarbonCoreDataHolder.getInstance().setBundleContext(bundleContext);
     }
 
--- a/components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/internal/Activator.java
+++ b/components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/internal/Activator.java
@@ -45,6 +45,9 @@ public class Activator implements BundleActivator {
 
     @Override
     public void start(BundleContext bundleContext) throws Exception {
+        if (log.isDebugEnabled()) {
+            log.debug(Activator.class.getName() + "#start() BEGIN - " + System.currentTimeMillis());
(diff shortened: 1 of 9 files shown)


components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/internal/Activator.java
```diff
@@ -45,6 +45,9 @@
 
     @Override
     public void start(BundleContext bundleContext) throws Exception {
+        if (log.isDebugEnabled()) {
+            log.info(Activator.class.getName() + "#start() BEGIN - " + System.currentTimeMillis());
```
Nirothipan: Is this going to be info or debug logs .

Nirothipan: Is this log.debug ?

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/internal/CoreServerInitializer.java
```diff
@@ -559,21 +529,21 @@ private void setAxis2RepoLocation() {
...
 
         if (!axis2RepoLocation.endsWith("/")) {
-            serverConfig
+            serverConfigurationService
                     .setConfigurationProperty(CarbonBaseConstants.AXIS2_CONFIG_REPO_LOCATION, axis2RepoLocation + "/");
             axis2RepoLocation = axis2RepoLocation + "/";
         }
     }
 
     private void registerCarbonServlet(HttpService httpService, HttpContext defaultHttpContext)
```
Nirothipan: Seems we don't use this method

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/internal/CoreServerInitializer.java
```diff
@@ -612,8 +582,8 @@ private void registerCarbonServlet(HttpService httpService, HttpContext defaultH
     }
 
     private void initNetworkUtils(AxisConfiguration axisConfiguration) throws AxisFault, SocketException {
```
Nirothipan: We can remove this method call and its usage as it relates to mgt console

milindaperera: cannot remove completely since it init server hostname not only management host name

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/internal/Activator.java
milindaperera: it should be debug. Fixed it.
Thanks

PR #731, opened 2019-09-09
Refactor Code

RidmiR:


integration/cli-tests/src/test/java/EnvSetup.sh
```diff
@@ -42,21 +47,25 @@ function extractCompressArchive() {
...
 }
 
+setup(){
 #Setting up the CLI environment
-
+echo "Working Directory "
+echo pwd
 #Check if the cli build is available in the location
 DIR="../../../cmd/build"
+pwd
```
Nirothipan: we don't need this line , isn't it ?

integration/cli-tests/src/test/java/EnvSetup.sh
```diff
@@ -42,21 +47,25 @@ function extractCompressArchive() {
...
     VERSION=$(cat pom.xml | grep "^    <version>.*</version>$" | awk -F'[><]' '{print $3}');
-    echo $VERSION
+    echo "Version : $VERSION"
 }
 
+setup(){
 #Setting up the CLI environment
-
+echo "Working Directory "
+echo pwd
```
Nirothipan: Shall we make these two to one line

PR #807, opened 2019-09-20
Add woodstox to classpath

milindaperera:
Originally this was identified when testcase fail org.wso2.carbon.esb.passthru.transport.test.PartialInputStreamReadError (Test case for Jira ESBJAVA-4616)

### DIFF
--- a/distribution/src/assembly/bin.xml
+++ b/distribution/src/assembly/bin.xml
@@ -541,6 +541,11 @@
                 <include>net.sf.saxon:Saxon-HE:jar</include>
                 <include>xerces.wso2:xercesImpl</include>
                 <include>commons-logging:commons-logging</include>
+                <!--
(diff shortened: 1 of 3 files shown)


distribution/src/assembly/bin.xml
```diff
@@ -299,6 +299,17 @@
...
         </fileSet>
+        <fileSet>
+            <directory>target/wso2carbon-core-${carbon.kernel.version}/lib</directory>
+            <outputDirectory>wso2mi-${pom.version}/wso2/lib</outputDirectory>
+            <includes>
+                <!--
+                    woodstox should be in class path to xml builder (need javax.xml.stream.XMLInputFactory
+                        implementation to axis2)
+                -->
+                <include>woodstox-core-asl-4.4.1.jar</include>
```
Nirothipan: Shall we add this as a mvn dependency and pack

PR #823, opened 2019-09-23
refactor(cli): update listing of artifacts

menuka94:
## Purpose
Refactor 'show' subcommands of all commands (related to MI artifacts) so that the printing logic is decoupled from the command's logic. In the existing implementation, code for printing tables is dependent on every command i.e. the following lines of code were duplicated in every 'show' command.
```go
                table := tablewriter.NewWriter(os.Stdout)	
		table.SetAlignment(tablewriter.ALIGN_LEFT)
		table.SetBorder(false)	
		table.SetColumnSeparator("  ")	
		table.Render()

```

With this change, an update to the table output can be introduced in one place such that it will be reflected across all commands. Now every 'show' command calls the `utils.PrintItemList()` function which handles the table-printing logic.

Additionally, Go `structs` related to artifacts (which were formerly in `structs.go`) are moved to separate files to improve the clarity of code.

This PR does not affect the functionality of the CLI.


milindaperera:
Let's create git hub issues for this improvements and doc changes

cmd/cmd/compositeAppShow.go
```diff
@@ -95,8 +96,8 @@ func executeGetCarbonAppCmd(appname string) {
 
 // Print the details of a Carbon app
 // Name, Version, and summary about it's artifacts
-// @param app : CarbonApp object
-func printCarbonAppInfo(app utils.CarbonApp) {
+// @param app : CompositeApp object
+func printCarbonAppInfo(app artifactUtils.CompositeApp) {
```
sdkottegoda: Any reason for not using the common function instead of this?

cmd/cmd/dataServiceShow.go
```diff
@@ -99,7 +100,7 @@ func printShowDataServiceHelp() {
 		"[data-service-name]") + showDataServiceCmdExmaples + utils.GetCmdFlags(dataServicesCmdLiteral))
 }
 
-func printDataServiceInfo(dataServiceInfo utils.DataServiceInfo) {
+func printDataServiceInfo(dataServiceInfo artifactUtils.DataServiceInfo) {
```
sdkottegoda: Any reason for not having a list command for dataservices? @sachithKay

cmd/cmd/localEntryShow.go
```diff
@@ -53,7 +52,7 @@ func init() {
 	localEntryCmd.AddCommand(localEntryShowCmd)
 	localEntryShowCmd.SetHelpTemplate(showLocalEntryCmdLongDesc +
 		utils.GetCmdUsage(programName, localEntryCmdLiteral,
-		utils.ShowCommand, "[localentry-name]") + showLocalEntryCmdExamples +
+			utils.ShowCommand, "[localentry-name]") + showLocalEntryCmdExamples +
```
sdkottegoda: Let's correct formatting :)

cmd/cmd/endpointShow.go
```diff
@@ -96,7 +95,7 @@ func executeGetEndpointCmd(endpointname string) {
 // Print the details of an Endpoint
 // Name, Type, Method, Url, Stats
 // @param Endpoint : Endpoint object
-func printEndpoint(endpoint utils.Endpoint) {
+func printEndpoint(endpoint artifactUtils.Endpoint) {
 
 	fmt.Println("Name - " + endpoint.Name)
```
sdkottegoda: Are the formats from different artifact types different? I don't see data being appended to a table as in APIs. @sajithaliyanage any idea?

cmd/cmd/localEntryShow.go
menuka94: I believe the formatting is correct as I ran [`go fmt`](https://blog.golang.org/go-fmt-your-code) on the project.

PR #650, opened 2019-08-23
Fix secure vault password length issue

sajithaliyanage:
## Purpose
Fix the issue of secure vault password length - https://github.com/wso2/micro-integrator/issues/593

PR #841, opened 2019-09-24
Add Null check for secureVaultRegex

Ibaqu:
+ This is required as setting null to a property throws an error

Related Issue : https://github.com/wso2/micro-integrator/pull/650

### DIFF
--- a/components/org.wso2.micro.integrator.security/src/main/java/org/wso2/micro/integrator/security/vault/VaultTool.java
+++ b/components/org.wso2.micro.integrator.security/src/main/java/org/wso2/micro/integrator/security/vault/VaultTool.java
@@ -234,6 +234,8 @@ public class VaultTool {
         }
 
         String secureVaultRegex = properties.getProperty(Constants.SECURE_VALULT_PASSWORD_REGEX_SYSTEM_PROPERTY);
-        System.setProperty(Constants.SECURE_VALULT_PASSWORD_REGEX_SYSTEM_PROPERTY, secureVaultRegex);
+        if (secureVaultRegex != null) {
+            System.setProperty(Constants.SECURE_VALULT_PASSWORD_REGEX_SYSTEM_PROPERTY, secureVaultRegex);
+        }
     }
 }


components/org.wso2.micro.integrator.security/src/main/java/org/wso2/micro/integrator/security/vault/VaultTool.java
```diff
@@ -234,6 +234,8 @@ private static void addSystemProperties() {
         }
 
         String secureVaultRegex = properties.getProperty(Constants.SECURE_VALULT_PASSWORD_REGEX_SYSTEM_PROPERTY);
-        System.setProperty(Constants.SECURE_VALULT_PASSWORD_REGEX_SYSTEM_PROPERTY, secureVaultRegex);
+        if (secureVaultRegex.isEmpty()) {
+            System.setProperty(Constants.SECURE_VALULT_PASSWORD_REGEX_SYSTEM_PROPERTY, secureVaultRegex);
+        }
     }
```
dilee: Could 'secureVaultRegex' ever be null?

Ibaqu: Yes. What happens is it gets the value from the carbon.property file first. 
If the property doesnt exist in the file it ends up as null and System.setProperty will throw a null pointer exception when it tries to set the property. 


With the null check is done, then it skips setting the property as null. And it uses the default regex in the previous function

Ibaqu: The previous function mentioned is here: https://github.com/wso2/micro-integrator/pull/650/files#diff-d9ed61ef411ddd14fd475144f74353ecR132 

This sets the proper regex if it cant find the property in the property file. 
```
String passwordRegex = System.getProperty(Constants.SECURE_VALULT_PASSWORD_REGEX_SYSTEM_PROPERTY);
if (Objects.isNull(passwordRegex)) {  
     System.out.println(Constants.SECURE_VALULT_PASSWORD_REGEX_SYSTEM_PROPERTY + " property cannot be found in the carbon.properties file. Hence setting the default value as " + Constants.SECURE_VALULT_PASSWORD_REGEX_SYSTEM_PROPERTY + "=^[\\\\S]{5,30}$\n");
}
```

components/org.wso2.micro.integrator.security/src/main/java/org/wso2/micro/integrator/security/vault/VaultTool.java
```diff
@@ -234,6 +234,8 @@ private static void addSystemProperties() {
         }
 
         String secureVaultRegex = properties.getProperty(Constants.SECURE_VALULT_PASSWORD_REGEX_SYSTEM_PROPERTY);
-        System.setProperty(Constants.SECURE_VALULT_PASSWORD_REGEX_SYSTEM_PROPERTY, secureVaultRegex);
+        if (secureVaultRegex.isEmpty()) {
```
sdkottegoda: We need to check add a null check instead of an empty check

Ibaqu: Done.

components/org.wso2.micro.integrator.security/src/main/java/org/wso2/micro/integrator/security/vault/VaultTool.java
Ibaqu: Done. I removed the check for empty. In hindsight its not really required. My mistake. 
It is now just checking for null. 

The only requirement for isEmpty would be if the property is set without a regex value. It will simply state that the password does not match the regular expression.

PR #853, opened 2019-09-25
update template files for configurations

madushadhanushka:
## Purpose
Add missing configurations and remove carbon class names

### DIFF
--- a/distribution/src/resources/config-tool/deployment-full.toml
+++ b/distribution/src/resources/config-tool/deployment-full.toml
@@ -28,15 +28,125 @@ internal_crypto_provider = "org.wso2.micro.integrator.crypto.provider.KeyStoreBa
 key = "SECRET_KEY" # inferred
 
 [super_admin]
+#usr-mgt.xml
+username = "admin"              # inferred
+password = "admin"              # inferred
+admin_role = "admin"            # inferred
+create_admin_account = true     # inferred
 
-username = "admin"
-password = "admin"
-create_admin_account = true
+[realm_manager]
+#usr-mgt.xml
+data_source = "WSO2CarbonDB"       # inferred
+properties.isCascadeDeleteEnabled = true    #inferred
+
+[user_store]
+#usr-mgt.xml
+class = "org.wso2.micro.integrator.security.user.core.ldap.ReadOnlyLDAPUserStoreManager" # inferred
+connection_url = "ldap://localhost:10389"   #inferred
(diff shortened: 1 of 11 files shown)


distribution/src/resources/config-tool/deployment-full.toml
```diff
@@ -156,14 +245,14 @@ truststore.type = "$ref{truststore.type}"
 truststore.password = "$ref{truststore.password}"
 
 ssl_profile.file_path= "conf/sslprofiles/listenerprofiles.xml"
-ssl_profile.read_interval = "30s"
+ssl_profile.read_interval = 3000       # interval in milliseconds
```
sdkottegoda: 30s != 3000ms... What do we want?

distribution/src/resources/config-tool/templates/conf/security/cipher-text.properties.j2
```diff
@@ -0,0 +1,6 @@
+# By default, This file contains the secret alias names and the plain text passwords enclosed with '[]' brackets
+# In Production environments, It is recommend to replace these plain text password by the encrypted values. CipherTool can be used for it.
+
+{% for secret_name,secret_value in secrets.items()%}
+{{secret_name}}={{secret_value}}
+{% endfor %}
```
Nirothipan: Please add a new line , else this will create issue when creating installers.

distribution/src/resources/config-tool/templates/conf/user-mgt.xml.j2
```diff
@@ -0,0 +1,41 @@
+<!--
+  ~ Copyright WSO2, Inc. (http://wso2.com)
```
Nirothipan: Shall we update the license

Nirothipan:
Shall we pack these files with filter enabled , else we won't get filtered properties when building distributions eg:-https://github.com/wso2/micro-integrator/pull/853/files#diff-a8fa48ccf3cf797bd983d7986a698362L39

distribution/src/resources/config-tool/templates/conf/user-mgt.xml.j2
madushadhanushka: fixed

distribution/src/resources/config-tool/templates/conf/security/cipher-text.properties.j2
madushadhanushka: Fixed

Issue #180, opened 2019-05-09
Implement a proper mechanism to store the metadata related with registry files


arunans23:
**Description:**

Currently a file based registry system exists in Micro Integrator and it does not have the capabilities to store the metadata such as Media-type. A temporary fix was done in /wso2/carbon-mediation#1247 to load the file extension mapping in the code level. But a proper mechanism should be implemented to overcome this issue.

**Suggested Labels:**

Improvement

**Suggested Assignees:**

**Affected Product Version:**
MI 1.0.0

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Related Issues:**

https://github.com/wso2/micro-integrator/issues/175

PR #866, opened 2019-09-26
Fix Validate JSON Schema Test case

Nirothipan:
## Purpose

Include fixes from https://github.com/wso2/carbon-mediation/pull/1270 which fixes 
 wso2/micro-integrator#180

The metadata related to registry resource is stored in a separate file named <resource_file_name>.meta inside the ".metadata" dir which is located where the registry resource is.
At the moment media type is the only metadata stored.

Example:
```
|registry/
| ---- | governance/
| -------- | custom1/
| ------------ | checkJsScript.js
| ------------ | .metadata/
| ---------------- | checkJsScript.js.meta
```

Fixes ValidateJSONSchemaTestCase .


milindaperera:
This is improvement to file based registry, not fixing the test case, hence shall we update description with the improvement made

Nirothipan:
> This is improvement to file based registry, not only fixing the test case, hence shall we update description with the improvement made

done .

components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
```diff
@@ -220,6 +225,24 @@ public OMNode lookup(String key) {
...
+    private boolean lookupUtil(String key, URL url) {
+        if (url == null) {
+            handleException("Unable to create URL for target resource : " + key);
+        }
+
+        if ("file".equals(url.getProtocol())) {
+            try {
+                url.openStream();
+            } catch (IOException e) {
+                if (log.isDebugEnabled()) {
```
sdkottegoda: Should this be an error log?

Nirothipan: This is just registry lookup and if the resource is not there we are returning null , hence I don't think this need to be an error log.

sdkottegoda: but is it correct to say that an IOException would be thrown only for a missing file?

Nirothipan: added in 0ee8df54ff7fa587de65c6d09435573d5c8f59f2

PR #867, opened 2019-09-26
Update full profile

Nirothipan:

### DIFF
--- a/integration/pom.xml
+++ b/integration/pom.xml
@@ -820,6 +820,9 @@
             </modules>
         </profile>
         <profile>
+            <!--
+               The profile with which build runs in WSO2 Jenkins
+                    -->
             <id>full</id>
             <modules>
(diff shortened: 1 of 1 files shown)


integration/pom.xml
```diff
@@ -820,14 +820,17 @@
...
+                    -->
             <id>full</id>
             <modules>
                 <module>samples/product</module>
                 <module>automation-extensions</module>
                 <module>tests-common</module>
                 <module>mediation-tests</module>
                 <module>cli-tests</module>
-<!--                <module>dataservice-hosting-tests</module>-->
+                <module>dataservice-hosting-tests</module>
```
sdkottegoda: Shall we enable dss tests in the smoke profile aswell?

Nirothipan: They are not fixed yet and travis will fail for the PRs

Issue #868, opened 2019-09-26
Update docs about Injecting parameters through environment variables


malakaganga:
**Description:**
We should package artifacts that need to be deployed in the micro integrator (synapse configs, APIs, proxies, inbound-endpoints) when we are creating a docker image.

In the current implementation of the synapse library, we need to specify parameters that can be varied in different environments in the synapse configuration itself.
So in order to achieve dynamic nature, we need to inject these values to the docker container in dynamic manner.
And the proposed solution is to injecting these values through environment variables.
So the docs should be updated on this each PR related to this feature will be linked to this issue

**Related Issues:**
https://github.com/wso2/micro-integrator/issues/847

Issue #847, opened 2019-09-24
Injecting parameters through environment variables into synapse configs


malakaganga:
**Description:**
We should package artifacts that need to be deployed in the micro integrator (synapse configs, APIs, proxies, inbound-endpoints) when we are creating a docker image.

In the current implementation of the synapse library, we need to specify parameters that can be varied in different environments in the synapse configuration itself. 
So in order to achieve dynamic nature, we need to inject these values to the docker container in dynamic manner.
And the proposed solution is to injecting these values through environment variables.

Following is a list of properties which will be require to parameterize

![Screenshot from 2019-09-24 18-26-20](https://user-images.githubusercontent.com/8380732/65513511-fed2e300-def8-11e9-8a71-fbd0cb0228c4.png)

**Affected Product Version:**
MI-1.1.0

**Related Issues:**
https://github.com/wso2/product-ei/issues/2735

PR #869, opened 2019-09-26
Parameterize data services related properties

malakaganga:
## Purpose
Provide the ability to resolve the following properties

1. URL
2. Username
3. Password

From environment variables when deploying inline/separate
data sources.

Related to: https://github.com/wso2/micro-integrator/issues/847

Docs: https://github.com/wso2/micro-integrator/issues/868

### DIFF
--- a/components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/resolvers/ResolverFactory.java
+++ b/components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/resolvers/ResolverFactory.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2018, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/resolvers/ResolverFactory.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private void registerResolvers() {
+        resolverMap.put("system", SystemResolver.class);
+    }
+
+    private void registerExterns() {
+        ServiceLoader<Resolver> loaders = ServiceLoader.load(Resolver.class);
+        for (Resolver resolver : loaders) {
+            String className = resolver.getClass().getName();
```
sameeragunarathne: I think we can use resolver.getClass().getSimpleName()

malakaganga: Fixed

Issue #254, opened 2019-06-06
[ DSS ] Support ODATA for Data Services In Micro Integrator.


Nirothipan:
**Description:**
$subject.

This was facilitated via odata.war file in EI. Since the web app deployment is not available in MI we need to implement an alternative.

Nirothipan:
https://docs.wso2.com/display/EI650/Exposing+Data+as+an+OData+Service

Nirothipan:
Few Integration tests are commented out due to this in MI. Please un comment them once fixed.

PR #875, opened 2019-09-26
Add OData testcases to MI

chanikag:
## Purpose
This PR is to add Odata Integration tests in MI.
Related issue : https://github.com/wso2/micro-integrator/issues/254


integration/dataservice-hosting-tests/tests-integration/tests/src/test/java/org/wso2/ei/dataservice/integration/test/odata/ODataETagTestCase.java
```diff
@@ -282,16 +276,16 @@ public void validateETagConcurrentHandlingTestCaseForDeleteMethod() throws Excep
...
-        String content = "{\"FILENAME\": \"M.K.H.Gunasekara\" ,\"TYPE\" : \"dss\"}";
+        String content = "{\"FILENAME\": \"WSO2PROD\" ,\"TYPE\" : \"dss\"}";
         Map<String, String> headers = new HashMap<>();
         headers.put("Accept", "application/json");
         Object[] response = sendPOST(endpoint, content, headers);
         Assert.assertEquals(response[0], ODataTestUtils.CREATED);
-        String entityEndpoint = webAppUrl + "/odata/" + serviceName + "/" + configId + "/FILES(\'M.K.H.Gunasekara\')";
-        response = sendGET(endpoint, headers);
+        String entityEndpoint = webAppUrl + "/odata/" + serviceName + "/" + configId + "/FILES(\'WSO2PROD\')";
+        response = sendGET(entityEndpoint, headers);
```
sdkottegoda: We have sent the request to **endpoint** earlier. Now we are sending it to **entityEndpoint**. Is this intentional?

chanikag: Yes. The idea of this testcase is to get etag of this particular field. But eventhough the url is set to filter this field, the previously defined url is used to invoke which gives all the fields. 

Earlier, the database scripts were ran per testcase. In that case, only the table is created and no rows are inserted. Therefore, as only one row is inserted before this line, only that field is getting retrieved. So the extracted etag matches exactly to this field. Therefore, this didn't fail. 

But now, we create the database after running all the scripts before running the test suite. Therefore, at this point there are multiple rows already available in the table. So the extracted etag does not exactly matches with this row.

PR #883, opened 2019-09-27
Refactor enviornment setup script

RidmiR:
## Purpose
FIx the CLI test fauiler issue observed in Jenkins. Refactor the code by adding the extracting the compressed file if its existence in the build path already.

### DIFF
--- a/integration/cli-tests/src/test/java/EnvSetup.sh
+++ b/integration/cli-tests/src/test/java/EnvSetup.sh
@@ -46,6 +46,7 @@ DIR="../../../../../cmd/build"
 
 if [ -d "$DIR" ]; then
     echo "CLI build exists."
+    cd ../../../../../cmd
 else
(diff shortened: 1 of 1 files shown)


integration/cli-tests/src/test/java/EnvSetup.sh
```diff
@@ -44,7 +44,8 @@ cd $BASEDIR
 DIR="../../../../../cmd/build"
 
 if [ -d "$DIR" ]; then
-    echo "CLI build exists. Hence skipping the environment setup phase"
+    echo "CLI build exists. Extracting Compress Archive"
+    extractCompressArchive
```
Nirothipan: shall we move this also outside if block, since it is there in if and else path

RidmiR: Done. Please review the latest commit

PR #928, opened 2019-09-30
Remove cli module from the full profile

RidmiR:
## Purpose
Remove cli module from the full profile


sdkottegoda:
Closing since we do not need to remove CLI tests

PR #947, opened 2019-10-01
Add reverseProxyMode property to passthru-http.properties and to deployment-full.toml files

sajinieKavindya:
### Purpose

- Add reverseProxyMode=false to passthru-http.properties
- Add reverse_proxy_mode = false to deployment-full.toml
- "transport.http.reverse_proxy_mode": "passthru_properties.'force.json.message.reverseProxyMode'" to key-mappings.json
- "passthru_properties.'force.json.message.reverseProxyMode'": false to default.json


Nirothipan:
Shall we create a doc issue or PR for this please

PR #967, opened 2019-10-02
Add missing templates

madushadhanushka:
Add missing templates for blocking transport


distribution/src/resources/config-tool/deployment-full.toml
```diff
@@ -73,6 +73,9 @@ read_timeout = '' # empty by default
 retry_attempts = '' # empty by default
 replace_escape_characters_at_user_login = true   #inferred
 
+[authorization_manager]
+class = "org.wso2.carbon.user.core.authorization.JDBCAuthorizationManager"
```
sdkottegoda: The package name is wrong. It should be "org.wso2.micro.integrator.security.user.core.authorization"

distribution/src/resources/config-tool/default.json
```diff
@@ -49,6 +49,11 @@
   "user_store.properties.LDAPConnectionTimeout": "5000",
   "user_store.properties.ReplaceEscapeCharactersAtUserLogin": true,
 
+  "authorization_manager.class": "org.wso2.carbon.user.core.authorization.JDBCAuthorizationManager",
```
sdkottegoda: The package name is wrong.

PR #969, opened 2019-10-02
Modify KeyStore files

sajinieKavindya:
## Purpose
Modify KeyStore files in tests-service module.


sdkottegoda:
Closing since this is not required.

PR #1028, opened 2019-10-04
Carbon log reader

sajinieKavindya:
## Purpose
Disable all the integration tests that use carbonLogReader


sdkottegoda:
Closing since this is not required anymore with the stabilized tescases.

PR #1148, opened 2020-02-13
Add readiness probe implementation.

GDLMadushanka:
Fixes wso2/micro-integrator/issues/1096


components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.probes/src/main/java/org/wso2/micro/integrator/probes/ReadinessResource.java
```diff
@@ -0,0 +1,94 @@
...
+        // Sending a response body for a GET request
+        axisCtx.removeProperty(NO_ENTITY_BODY);
+        // appending MI server version number to the response
+        CarbonServerConfigurationService serverConfig = CarbonServerConfigurationService.getInstance();
+        String miVersion = serverConfig.getServerVersion();
+        String response = "{\"version\":\"" + miVersion + "\",";
+
+        ArrayList<String> faultyCapps = CAppDeploymentManager.getFaultyCapps();
+        if (faultyCapps.size() > 0) {
+            response += "\"status\": \"not ready, faulty CAPPs detected\"}";
```
Nirothipan: will that be useful if we append the list of fault caps to the response.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.probes/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Bundle-Name>${project.artifactId}</Bundle-Name>
+                        <Export-Package>
+                            org.wso2.micro.integrator.probes.*; version="${project.version}"
+                        </Export-Package>
+                        <Import-Package>
+                            *;resolution:=optional
```
Nirothipan: Shall we list the known imports here

Nirothipan:
Shall we add an integration test as well for this.

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
```diff
@@ -108,8 +111,11 @@ public static void loadInternalApis(String apiFilePath) {
                             if (name == null || name.isEmpty()) {
                                 handleException("Name not specified in one or more handlers");
                             }
+                            // Readiness-probe api is enabled by default.
+                            // other APIs must be enabled manually.
                             if (!Boolean.parseBoolean(
-                                    System.getProperty(Constants.PREFIX_TO_ENABLE_INTERNAL_APIS + name))) {
+                                    System.getProperty(Constants.PREFIX_TO_ENABLE_INTERNAL_APIS + name)) &&
+                                    !READINESS_PROBE_API_NAME.equals(name)) {
```
Nirothipan: Instead of enabling here what if we have this system property in the start up script. If someone doesn't want to expose this ( as this is over http and have no security ) they can remove the system property in script but if we enable here they won't be able to restrict it easily.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.probes/src/main/java/org/wso2/micro/integrator/probes/ReadinessResource.java
```diff
@@ -0,0 +1,94 @@
...
+        axisCtx.setProperty(Constants.Configuration.CONTENT_TYPE, "application/json");
+        // Sending a response body for a GET request
+        axisCtx.removeProperty(NO_ENTITY_BODY);
+        // appending MI server version number to the response
+        CarbonServerConfigurationService serverConfig = CarbonServerConfigurationService.getInstance();
+        String miVersion = serverConfig.getServerVersion();
+        String response = "{\"version\":\"" + miVersion + "\",";
+
+        ArrayList<String> faultyCapps = CAppDeploymentManager.getFaultyCapps();
+        if (faultyCapps.size() > 0) {
```
Nirothipan: Shall we add a configurable parameter here to return 200 even with faulty capss. i.e by default we will return 500 if fault but it can be configurable that we can return 200 if that system property is set to true. 
Then we can use this probe in Integration Studio as well to make sure that server is started ( currently we use Prometheus ) , in that case we can disable the system property and let server start with faulty capss.

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
abeykoon: Irrespective of the fact that internal APIs are enabled or not, we need to run the readiness probe.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.probes/src/main/java/org/wso2/micro/integrator/probes/ReadinessResource.java
abeykoon: Have we tested the scenarios where we have multiple CAPPs and only one has failed? 
What if a particular CAPP takes a long time to deploy? Does API wait until that CAPP finish its deployment to return 200?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.probes/src/main/java/org/wso2/micro/integrator/probes/ReadinessResource.java
abeykoon: I suppose it is useful if we can return what are the faulty apps are.

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
Nirothipan: yeah but we can make it configurable. That is we can define the system property in startup script without forcing it in the code level so that people who doesn't want to have the probe can stop it.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.probes/src/main/java/org/wso2/micro/integrator/probes/ReadinessResource.java
GDLMadushanka: +1 fixed in 53121f0b023e309f3de70ac5a12dbc95616e479d

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
GDLMadushanka: +1 Fixed in 53121f0b023e309f3de70ac5a12dbc95616e479d

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.probes/src/main/java/org/wso2/micro/integrator/probes/ReadinessResource.java
GDLMadushanka: If the purpose is only to check the server is started, tooling can just check the response payload != null.

Nirothipan: AFAIK the server plugin used in tooling uses this service and it needs a 200 status code.

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
```diff
@@ -108,8 +111,12 @@ public static void loadInternalApis(String apiFilePath) {
                             if (name == null || name.isEmpty()) {
                                 handleException("Name not specified in one or more handlers");
                             }
+                            // Readiness-probe api is enabled by default. ( to disable use -DdisableReadiness=true )
+                            // other APIs must be enabled manually.
                             if (!Boolean.parseBoolean(
-                                    System.getProperty(Constants.PREFIX_TO_ENABLE_INTERNAL_APIS + name))) {
+                                    System.getProperty(Constants.PREFIX_TO_ENABLE_INTERNAL_APIS + name)) &&
+                                    !(READINESS_PROBE_API_NAME.equals(name) && !Boolean.parseBoolean(
```
Nirothipan: Instead of complicating here , shall we just add to startup script without touching this code base as discussed offline

integration/mediation-tests/tests-other/src/test/resources/testng.xml
```diff
@@ -121,9 +121,9 @@
         </classes>
     </test>
 
-    <test name="LocalEntry-Test" preserve-order="true" verbose="2">
+    <test name="Kubernetes probe tests" preserve-order="true" verbose="2">
         <packages>
-            <package name="org.wso2.carbon.esb.localentry.test.*"/>
```
Nirothipan: Are we deleting local entry test cases here

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.probes/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.wso2.ei</groupId>
+            <artifactId>org.wso2.micro.integrator.inbound.endpoint</artifactId>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.wso2.ei</groupId>
+            <artifactId>org.wso2.micro.integrator.initializer</artifactId>
```
sdkottegoda: Is it correct for extensions to have a dependency to the initializer? Would this introduce a cyclic dependency?

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/services/CarbonServerConfigurationService.java
```diff
@@ -552,4 +557,13 @@ protected boolean isProtectedToken(String key) {
...
 		return secretResolver.resolve("Carbon." + key);
 	}
+
+	/**
+	 * Take the WSO2 server version from the carbon.xml file.
+	 *
+	 * @return WSO2 server version.
+	 */
+	public String getServerVersion() {
+		return getFirstProperty(SERVER_VERSION);
```
sdkottegoda: Shall we correct the indentation?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.probes/src/main/java/org/wso2/micro/integrator/probes/ReadinessResource.java
```diff
@@ -0,0 +1,110 @@
...
+     */
+    @Override
+    public boolean invoke(MessageContext synCtx) {
+        org.apache.axis2.context.MessageContext axisCtx = ((Axis2MessageContext) synCtx).getAxis2MessageContext();
+        axisCtx.setProperty(Constants.Configuration.MESSAGE_TYPE, "application/json");
+        axisCtx.setProperty(Constants.Configuration.CONTENT_TYPE, "application/json");
+        // Sending a response body for a GET request
+        axisCtx.removeProperty(NO_ENTITY_BODY);
+
+        String response = "{\"version\":\"";
```
sdkottegoda: Can we initialize response inside the else block. I think it improves readability, WDYT?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/deployment/application/deployer/CAppDeploymentManager.java
```diff
@@ -376,5 +375,12 @@ private void handleException(String msg, Exception e) throws CarbonException {
         throw new CarbonException(msg, e);
     }
 
-
+    /**
+     * Get a list of faulty CAPPs in the server.
+     *
+     * @return list of faulty CAPPs
+     */
+    public static ArrayList<String> getFaultyCapps() {
```
sdkottegoda: It's better if we can return an unmodifiable list

Issue #1049, opened 2019-10-17
Add correlation logger and appender to log4j2.properties file


chanikag:
**Description:**
Currently user has to follow following steps,

1.  Add the following parameters to the log4j2.properties file (stored in the `MI_HOME/conf/` directory):

```
    # Appender config to put correlation Log.
    appender.CORRELATION.type = RollingFile
    appender.CORRELATION.name = CORRELATION
    appender.CORRELATION.fileName = ${sys:carbon.home}/repository/logs/correlation.log
    appender.CORRELATION.filePattern = ${sys:carbon.home}/repository/logs/correlation-%d{MM-dd-yyyy.log
    appender.CORRELATION.layout.type = PatternLayout
    appender.CORRELATION.layout.pattern = %d{yyyy-MM-dd HH:mm:ss,SSS}|%X{Correlation-ID}|%t|%m%n
    appender.CORRELATION.policies.type = Policies
    appender.CORRELATION.policies.time.type = TimeBasedTriggeringPolicy
    appender.CORRELATION.policies.time.interval = 1
    appender.CORRELATION.policies.time.modulate = true
    appender.CORRELATION.policies.size.type = SizeBasedTriggeringPolicy
    appender.CORRELATION.policies.size.size = 10MB
    appender.CORRELATION.strategy.type = DefaultRolloverStrategy
    appender.CORRELATION.strategy.max = 20
    appender.CORRELATION.filter.threshold.type = ThresholdFilter
    appender.CORRELATION.filter.threshold.level = INFO 
```
     
There are some differences with the existing configs

2.  Add `correlation` as a logger
    `loggers = AUDIT_LOG, SERVICE_LOGGER, correlation,`
    
3.  Add `CORRELATION` as an appender
   ` appenders = CARBON_CONSOLE, CARBON_LOGFILE, AUDIT_LOGFILE, CORRELATION, `

It is better if these are added by default to the log4j2.properties file. 

Related documentation: https://ei.docs.wso2.com/en/latest/micro-integrator/administer-and-observe/observability/

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Related Issues:**

PR #1151, opened 2020-02-17
Update log4j2 file with correlation logs

Nirothipan:
Add the correlation properties by default to log4j2 prop file so that we can simply enable
correlation logs using system properties and view

Fixes https://github.com/wso2/micro-integrator/issues/1049


distribution/src/conf/log4j2.properties
```diff
@@ -191,7 +192,7 @@ loggers = AUDIT_LOG, SERVICE_LOGGER, trace-messages, org-apache-coyote, com-haze
   BlueprintExtender, ModuleDeployer, VFSTransportSender, PassThroughHttpSender, PassThroughHttpSSLSender, \
   PassThroughHttpListener, PassThroughHttpSSLListener, DeploymentEngine, TaskServiceImpl, SynapseControllerFactory, \
   Axis2SynapseController, MultiXMLConfigurationBuilder, XMLConfigurationBuilder, \
-  VFSTransportListener, NTaskTaskManager, MediationStatisticsComponent, SynapseConfigurationBuilder, ServerManager
+  VFSTransportListener, NTaskTaskManager, MediationStatisticsComponent, SynapseConfigurationBuilder, ServerManager, correlation
```
sdkottegoda: Do we need correlation logs enabled by default? Also, let's add this into the below section with verbose logging aswell?

Nirothipan: Since we use a system property to enable the co-relation logs, having this by default won't enable it. It is added by default to reduce the config changes user has to do to enable it.

sdkottegoda: Let's add it to the verbose section aswell.

PR #1184, opened 2020-02-27
Fix issue in updating log4j2 properties log level using MI CLI

Vathsan:


sdkottegoda:
Shall we add the issue that was resolved?

Vathsan:
> Shall we add the issue that was resolved?

Updated the description

PR #1214, opened 2020-03-10
Remove property map from coordination tables

Nirothipan:
$subject since it is not used.


components/org.wso2.micro.integrator.coordination/src/main/java/org/wso2/micro/integrator/coordination/RDBMSCoordinationStrategy.java
```diff
@@ -341,22 +327,15 @@ public void run() {
...
-
         /**
          * Constructor.
          *
          * @param nodeId  node ID of the current node
          * @param groupId group ID of the current group
          */
-        private CoordinatorElectionTask(String nodeId, String groupId, Map<String, Object> propertiesMap,
-                                        boolean stillCoordinator) {
+        private CoordinatorElectionTask(String nodeId, String groupId, boolean stillCoordinator) {
```
sajinieKavindya: Shall we add the stillCoordinator parameter to the JavaDoc?

PR #1217, opened 2020-03-12
Add method to set hot deployment parameter value

sajinieKavindya:
## Purpose
Before this PR, the default value of the hotDeployment parameter was being taken as true by default during the server startup time and the value was being read from axis configuration using the following code line. 

`Parameter hotDeployment = axisConfiguration.getParameter("hotdeployment");`

Other than that, in some places `isHotDeploymentEnabled()` method inside MicroIntegratorBaseUtils class was being used which will read the axis2.xml file directly and return the value of the hot deployment.

Having two different methods and considering their implementations, this may lead to two contradicting behavior in the server since the two methods can give contradicting values for hot deployment at the same time. (one method can give true as a result while other one is giving false at the same time).

Therefore to address the above issues and to reduce the code redundancy, this PR has created a reference to hot deployment in the CarbonAxisConfigurator class and access it via a util class whenever needed. This will also make the hotDeployment as false by default.

### DIFF
--- a/components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/core/CarbonAxisConfigurator.java
+++ b/components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/core/CarbonAxisConfigurator.java
@@ -46,6 +46,7 @@ import org.apache.axis2.description.Parameter;
 import org.apache.axis2.engine.AxisConfiguration;
 import org.apache.axis2.engine.AxisConfigurator;
 import org.apache.axis2.i18n.Messages;
+import org.apache.axis2.util.JavaUtils;
 import org.apache.axis2.util.XMLUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -86,6 +87,7 @@ public class CarbonAxisConfigurator extends DeploymentEngine implements AxisConf
     }
 
     public CarbonAxisConfigurator() {
+        this.hotDeployment = false;
     }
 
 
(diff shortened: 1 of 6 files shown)


components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/core/CarbonAxisConfigurator.java
```diff
@@ -238,6 +241,26 @@ public boolean isGlobalyEngaged(AxisModule axisModule) {
         return false;
     }
 
+    /**
+     * Set the hotDeployment variable after reading the hot deployment parameter from the axis configuration. If the
+     * value is null, then the hotDeployment variable will be set to false which is the default value.
+     */
+    private void setHotDeploymentFeature() {
```
Vathsan: Shall we change this method name to updateHotDeploymentParameter? WDYT?

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/core/CarbonAxisConfigurator.java
```diff
@@ -238,6 +241,26 @@ public boolean isGlobalyEngaged(AxisModule axisModule) {
...
+        Parameter hotDeployment = axisConfig.getParameter(org.wso2.micro.integrator.core.Constants.HOT_DEPLOYMENT);
+        if (hotDeployment != null) {
```
Vathsan: As isTrue() method can handle null value and will return false, do we need this null check?

sajinieKavindya: Yes. `isTrue()` is evaluating `hotDeployment.getValue()` (not the `hotDeployment` value itself). The null check is there to avoid any NullPointerExceptions when getting `hotDeployment.getValue()`.

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/core/CarbonAxisConfigurator.java
sajinieKavindya: Updated.

Issue #1139, opened 2020-02-12
Remove CLI and Monitoring Dashboard code base from micro-integrator repo


Nirothipan:
**Description:**
$subject as this has been moved to https://github.com/wso2/product-mi-tooling.

CLI test cases need to be fixed as this is currently getting the CLI distribution locally and running the tests https://github.com/wso2/micro-integrator/tree/master/integration/cli-tests

PR #1225, opened 2020-03-13
Change CLI tests to download CLI from Jenkins

sdkottegoda:
## Purpose
Changes the CLI tests to download the CLI tool from the latest successful build of Jenkins so that the CLI code can be removed from this repository as explained in https://github.com/wso2/micro-integrator/issues/1139.

### DIFF
--- a/integration/cli-tests/src/test/java/EnvSetup.sh
+++ b/integration/cli-tests/src/test/java/EnvSetup.sh
@@ -26,56 +26,23 @@ BASEDIR=$(dirname "$0")
 echo "Platform : $platform"
 echo "BitType : $bitType"
 
-getPomVersion(){
-    VERSION=$(cat pom.xml | grep "^    <version>.*</version>$" | awk -F'[><]' '{print $3}');
-    echo "Version : $VERSION"
-}
-
-buildCli(){
-    go mod vendor
-    sleep 10
-
-    #build Micro Integrator CLI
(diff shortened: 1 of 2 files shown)


integration/cli-tests/src/test/java/EnvSetup.sh
```diff
@@ -26,56 +26,23 @@ BASEDIR=$(dirname "$0")
...
-
 #Setting up the CLI environment
 #Check if the cli build is available in the location
-setup(){
-
 cd $BASEDIR
 
-DIR="../../../../../cmd/build"
+#Get the CLI distribution from the latest jenkins build
+DISTRIBUTION_NAME=$(eval "curl -s https://wso2.org/jenkins/job/products/job/product-mi-tooling/lastSuccessfulBuild/api/json | python3 -c \"import sys, json; print(json.load(sys.stdin)['artifacts'][1]['fileName'])\"")
```
Nirothipan: This is going to run this only in linux,   If we can consider platform and bit type while downloading then we can run in others as well

sdkottegoda: I did not want to do it since the purpose of this change is different, let's consider it as a next step. WDYT?

Nirothipan: +1

PR #1227, opened 2020-03-15
Add coordination support for scheduled tasks in Micro Integrator

Nirothipan:
## Goals

> Provide the following capabilities for coordinated task deployment

- Schedule tasks in a single node in the cluster upon deployment.
- The task needs to be rescheduled in another node when the task running node becomes unavailable (HA).
- The state of the task needs to be maintained upon rescheduling ( if it was in a paused state ).
- Tasks need to be load balanced among the nodes in the cluster.
- Users need a task to be scheduled only in particular nodes ( pinned servers).

 

[Task Co-ordination.pdf](https://github.com/wso2/micro-integrator/files/4334034/Task.Co-ordination.pdf)

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes
 - Ran FindSecurityBugs plugin and verified report? yes
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes


components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/CoordinateTaskRunTimeException.java
```diff
@@ -0,0 +1,27 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.micro.integrator.ntask.coordination.task;
+
+public class CoordinateTaskRunTimeException extends RuntimeException {
```
sdkottegoda: Is it essential to make this a runtime exception?

Nirothipan: As discussed in the code review , this is used when there is a not supported db type configured.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/ClusterNodeDetails.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.List;
+
+/**
+ * The class which provides the cluster node details. No additional controlling methods should be added to this un
+ * necessarily as this is exposed to task location resolver. Task location will be resolved only in leader nodes.
+ *
+ * @see org.wso2.micro.integrator.ntask.coordination.task.resolver.TaskLocationResolver
+ */
+public class ClusterNodeDetails {
```
sdkottegoda: I feel like this name is a bit confusing. From the name along it seems like it doesn't belong to the ntask module and it wrapps a coordinator. Can't we have a better name like 'ClusterCoordinator' may be? 'NodeDetails' seems like just a class that has node details. But this would basically be our delegator for the org.wso2.micro.integrator.coordination.ClusterCoordinator.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/TaskDataBase.java
```diff
@@ -0,0 +1,157 @@
...
+ * The layer which connects to the task data base.
+ */
+public class TaskDataBase {
+
+    /**
+     * Connector for the data base.
+     */
+    private CoordinatedTaskDBConnector coordinatedTaskDBConnector;
+
+    public TaskDataBase(DataSource dataSource) throws SQLException {
```
sdkottegoda: Can we create an Exception wrapping the SQLException for that TaskDBCleaner doesn't have to know about the underlying sql implementation. Basically, if we decide to replace the data layer with some other type (file, nosql, etc.) we wouldn't have to change any code in TaskDBCleaner.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.sql.DataSource;
+
+/**
+ * The connector class which deals with underlying coordinated task table.
+ */
+public class CoordinatedTaskDBConnector {
```
sdkottegoda: I feel like it's better if we can rename this class to something generic like 'DBConnector' since this is just a dumb db query executor. This being inside the package says that it has something to do with tasks.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/ClusterNodeDetails.java
Nirothipan: ClusterCoordinator is fine but we have the same class in coordination as well . What if we use ClusterConnector ? or any other better suggestion

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
Nirothipan: If so we have give to pass the queries also. It is more coupled with TASK_TABLE , thats why I named it like it . wdyt

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/TaskDataBase.java
sdkottegoda: This wrappe exception needs to be thrown by the CoordinatedTaskDBConnector itself.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+            }
+        }
+    }
+
+    /**
+     * Removes the node id of the task and update the task state.
+     *
+     * @param coordinatedTasks - List of coordinated tasks which needs to be updated.
+     */
+    public void cleanInvalidNodesAndUpdateState(List<CoordinatedTask> coordinatedTasks) throws SQLException {
```
sdkottegoda: This class has no idea about invalid nodes. What it does it updating the state of a give task list to none. So shall we rename this to something like, "updateTaskStateToNone"

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+        }
+    }
+
+    /**
+     * For all the tasks which has this destined node id , sets it to null and update the task state to none if it was
+     * in a running state.
+     *
+     * @param nodeId - Node Id.
+     */
+    public void cleanTasksOfNode(String nodeId) throws SQLException {
```
sdkottegoda: Again let's be explicit about what this method does. Maybe 'unassignRunningTasksAndUpdateStateToNone'

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/TaskQueryHelper.java
```diff
@@ -0,0 +1,76 @@
...
+                    + ConstantUtil.TASK_STATE + " = ? WHERE " + ConstantUtil.TASK_NAME + " =?";
+
+    public static final String REMOVE_TASKS_OF_NODE =
+            "DELETE FROM " + ConstantUtil.TABLE_NAME + "  WHERE " + ConstantUtil.DESTINED_NODE_ID + " =?";
+
+    public static final String DELETE_TASK =
+            "DELETE FROM " + ConstantUtil.TABLE_NAME + " WHERE " + ConstantUtil.TASK_NAME + " =?";
+
+    public static final String CLEAN_TASKS_OF_NODE =
+            "UPDATE " + ConstantUtil.TABLE_NAME + " SET " + ConstantUtil.DESTINED_NODE_ID + " = NULL , "
```
sdkottegoda: Should we set the DESTINED_NODE_ID to null even when the task are not in a running state. e.g., if the task was completed, we might loose information as to who completed it, whether there was a bug, etc.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+            return;
+        }
+        try (Connection connection = getConnection(); PreparedStatement preparedStatement = connection.prepareStatement(
+                TaskQueryHelper.CLEAN_INVALID_NODES)) {
+            for (CoordinatedTask task : coordinatedTasks) {
+                String taskName = task.getTaskName();
+                preparedStatement.setString(1, task.getTaskState().name());
+                preparedStatement.setString(2, taskName);
+                preparedStatement.addBatch();
+                if (LOG.isDebugEnabled()) {
```
sdkottegoda: Shall we have the debug log after executing the statement as well? Would it be helpful for us in debugging issues?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+     * in a running state.
+     *
+     * @param nodeId - Node Id.
+     */
+    public void cleanTasksOfNode(String nodeId) throws SQLException {
+
+        try (Connection connection = getConnection(); PreparedStatement preparedStatement = connection.prepareStatement(
+                TaskQueryHelper.CLEAN_TASKS_OF_NODE)) {
+            preparedStatement.setString(1, nodeId);
+            LOG.debug("Cleaning the tasks of node [" + nodeId + "].");
```
sdkottegoda: Redundant debug log

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    /**
+     * Retrieves the list of tasks of node specified in specified state.
+     *
+     * @param nodeID - Id of the node, for which the tasks need to be retrieved.
+     * @param state  - State of the tasks which need to be retrieved.
+     * @return - List of tasks.
+     */
+    public List<String> retrieveTasksOfParticularNodeInSpecifiedState(String nodeID, CoordinatedTask.States state)
```
sdkottegoda: we don't have to say 'OfParticularNodeInSpecifiedState', passing the arguments tell that.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+            throws SQLException {
+
+        List<String> taskList = new ArrayList<>();
```
sdkottegoda: Why are we initializing this list even before creating the connection? I feel like it should be initialize inside the method 'populateCoordinatedTaskNames' WDYT?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+                TaskQueryHelper.RETRIEVE_TASKS_OF_PARTICULAR_NODE)) {
+            preparedStatement.setString(1, nodeID);
+            preparedStatement.setString(2, state.name());
+            populateCoordinatedTaskNames(preparedStatement, taskList);
+        }
+        debugLogs(taskList);
+        return taskList;
+    }
+
+    private void debugLogs(List<String> tasks) {
```
sdkottegoda: may be printDebugLogs?

sdkottegoda: If you introduced this method just because you cannot have the same name, the solution would be to override the toString() Method in CoordinatedTask class and to have generic object list as the input parameter.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+    private void debugLogs(List<String> tasks, String msg) {
+        if (LOG.isDebugEnabled()) {
+            LOG.debug(msg);
+            for (String task : tasks) {
+                LOG.debug(" Task Name [" + task + "].");
+            }
+        }
+    }
+
+    private void printDebugLogs(List<CoordinatedTask> tasks) {
```
sdkottegoda: We can remove this method and the below one if we have private void printDebugLogs(List<Object> tasks) {...}

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+            LOG.debug("Removed all the tasks of node [" + nodeId + "].");
+        }
+    }
+
+    /**
+     * Remove the task entry from the db.
+     *
+     * @param coordinatedTasks - List of tasks to be removed.
+     */
+    public void removeTasksFromDB(List<String> coordinatedTasks) throws SQLException {
```
sdkottegoda: the input parameter could be just tasks

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+
+        if (coordinatedTasks.isEmpty()) {
+            return;
```
sdkottegoda: Shall we add a debug log here?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    /**
+     * Helper method to populate coordinated tasks to the list.
+     *
+     * @param preparedStatement - Statement to be executed to retrieve the list of tasks.
+     * @param taskList          - The list to be populated.
+     * @throws SQLException - Exception.
+     */
+    private void populateCoordinatedTaskNames(PreparedStatement preparedStatement, List<String> taskList)
```
sdkottegoda: I wouldn't like mutating the task list that was passed. Rather creating a new object and returning from here.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    /**
+     * Helper method to populate coordinated tasks to the list.
+     *
+     * @param preparedStatement - Statement to be executed to retrieve the list of tasks.
+     * @param taskList          - The list to be populated.
+     * @throws SQLException - Exception.
+     */
+    private void populateCoordinatedTaskList(PreparedStatement preparedStatement, List<CoordinatedTask> taskList)
```
sdkottegoda: same as above

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+            }
+        }
+    }
+
+    /**
+     * Retrieve all the tasks in the task db.
+     *
+     * @return - List of available tasks.
+     */
+    public List<String> getAllTasksInDB() throws SQLException {
```
sdkottegoda: We are only returning taskNames, so shall we rename the method to imply that? Also we don;t need to say from DB, because this is the DBConnector.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/ConstantUtil.java
```diff
@@ -0,0 +1,42 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.micro.integrator.ntask.coordination.task;
+
+/**
+ * Constant class.
+ */
+public class ConstantUtil {
```
sdkottegoda: Why can't we just call this Constants? Because we do not have any util methods here?

sdkottegoda: Aslo do we need to classes, this and the TaskQueryHelper? Can we consolidate the constants to a single class?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/TaskQueryHelper.java
```diff
@@ -0,0 +1,76 @@
...
+                    + ConstantUtil.TASK_STATE + " = ( CASE " + ConstantUtil.TASK_STATE + " WHEN \""
+                    + CoordinatedTask.States.RUNNING + "\"THEN \"" + CoordinatedTask.States.NONE + "\" ELSE "
+                    + ConstantUtil.TASK_STATE + " END ) WHERE " + ConstantUtil.DESTINED_NODE_ID + " = ?";
+
+    public static final String GET_ALL_ASSIGNED_IN_COMPLETE_TASKS =
```
sdkottegoda: IN_COMPLETE? Or INCOMPLETE?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+        debugLogs(taskList);
+        return taskList;
+    }
+
+    /**
+     * Retrieve all assigned and in completed tasks in the task db.
+     *
+     * @return - List of available tasks.
+     */
+    public List<CoordinatedTask> getAllAssignedInCompleteTasks() throws SQLException {
```
sdkottegoda: InComplete or Incomplete?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+        printDebugLogs(taskList);
+        return taskList;
+    }
+
+    /**
+     * Add the task to the task db.
+     *
+     * @param coordinatedTask - The coordinated task which needs to be added.
+     */
+    public void addTaskToDB(CoordinatedTask coordinatedTask) throws SQLException {
```
sdkottegoda: ToDb is not necessary :)

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+
+        String taskName = coordinatedTask.getTaskName();
+        CoordinatedTask.States state = coordinatedTask.getTaskState();
+        try (Connection connection = getConnection(); PreparedStatement preparedStatement = connection.prepareStatement(
+                TaskQueryHelper.ADD_TASK)) {
+            preparedStatement.setString(1, taskName);
+            preparedStatement.setString(2, state.name());
+            preparedStatement.executeUpdate();
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Successfully added the task [" + taskName + "] with state [" + state
```
sdkottegoda: We can simply print the task object if we override the toString() method

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+            }
+        }
+    }
+
+    /**
+     * Updates the state and node id of the provided tasks in db.
+     *
+     * @param coordinatedTasks - List of tasks to be updated.
+     */
+    public void updateTaskDB(List<CoordinatedTask> coordinatedTasks) throws SQLException {
```
sdkottegoda: updatedTasks() may be?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
```diff
@@ -0,0 +1,362 @@
...
+
+        if (coordinatedTasks.isEmpty()) {
+            return;
+        }
+        try (Connection connection = getConnection(); PreparedStatement preparedStatement = connection.prepareStatement(
+                TaskQueryHelper.UPDATE_TASK)) {
+            for (CoordinatedTask coordinatedTask : coordinatedTasks) {
+                String taskName = coordinatedTask.getTaskName();
+                String nodeId = coordinatedTask.getDestinedNodeId();
+                CoordinatedTask.States state = coordinatedTask.getTaskState();
```
sdkottegoda: Do we need these assignements? doesn't really matter though

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
sdkottegoda: But it's inside the task package, so the DBConnector inside the task package gives you the idea that it deals with the tasks table, does it not?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/TaskQueryHelper.java
Nirothipan: This query will set the state to none only for running tasks not others.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/CoordinatedTaskDBConnector.java
Nirothipan: Thanks . Nice suggestion

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/ClusterCommunicator.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.List;
+
+/**
+ * The class which provides the cluster node details. No additional controlling methods should be added to this un
+ * necessarily as this is exposed to task location resolver. Task location will be resolved only in leader nodes.
+ *
+ * @see org.wso2.micro.integrator.ntask.coordination.task.resolver.TaskLocationResolver
+ */
+public class ClusterCommunicator {
```
sdkottegoda: +1

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/CoordinatedTaskException.java
```diff
@@ -0,0 +1,27 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.micro.integrator.ntask.coordination;
+
+public class CoordinatedTaskException extends Exception {
```
sdkottegoda: TaskCoordinationException?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/connector/RDMBSConnector.java
```diff
@@ -0,0 +1,403 @@
...
+                TaskQueryHelper.RETRIEVE_TASKS_OF_NODE)) {
+            preparedStatement.setString(1, nodeID);
+            preparedStatement.setString(2, state.name());
+            return query(preparedStatement, "for node [" + nodeID + "] with state [" + state.name() + "]");
+        } catch (SQLException ex) {
+            throw new CoordinatedTaskException(ERROR_MSG, ex);
+        }
+    }
+
+    private void debugLogs(List<Object> tasks, String msg) {
```
sdkottegoda: Shall we rename this to something like printDebugLogs?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/TaskQueryHelper.java
```diff
@@ -0,0 +1,81 @@
...
+    //task table Name
+    private static final String TABLE_NAME = "COORDINATED_TASK_TABLE";
+
+    // Task table columns
+    public static final String TASK_NAME = "TASK_NAME";
+    public static final String DESTINED_NODE_ID = "DESTINED_NODE_ID";
+    public static final String TASK_STATE = "TASK_STATE";
+
+    public static final String ADD_TASK_IF_NOT_EXISTS =
+            "INSERT IGNORE INTO " + TABLE_NAME + " ( " + TASK_NAME + ", " + DESTINED_NODE_ID + ", " + TASK_STATE + ") "
```
sdkottegoda: Does this work in mssql or other RDBMSs? Let's try to stick to the basic syntaxes so that it won't cause any issues when we add other DB types. WDYT?

Nirothipan: mssql have some other syntax. but there are should be equivalent syntax. we can fix them while testing (  set constants while initialization the db ), else we won't be able to do batch updates and would have to do one by one. Also we would have to catch data integrity violation each time and process that also. Any other better solution ?

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/util/MicroIntegratorBaseUtils.java
```diff
@@ -560,4 +564,21 @@ public static void setCarbonAxisConfigurator(CarbonAxisConfigurator carbonAxisCo
...
+     * Get Synapse Environment.
+     *
+     * @return SynapseEnvironment - SynapseEnvironment
+     */
+    public static SynapseEnvironment getSynapseEnvironment() {
+
+        Parameter synapseEnvironmentParatemer =
+                CarbonCoreDataHolder.getInstance().getAxis2ConfigurationContextService().getServerConfigContext()
+                        .getAxisConfiguration().getParameter(SynapseConstants.SYNAPSE_ENV);
+        if (synapseEnvironmentParatemer != null) {
```
sdkottegoda: Can this ever be null? Is this method getting called before the sunapseEnvironment is set?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/core/internal/TasksDSComponent.java
```diff
@@ -142,6 +192,11 @@ private Properties getStandardQuartzProps() {
     @Deactivate
     protected void deactivate(ComponentContext ctx) {
 
+        ScheduledExecutorService executorService = dataHolder.getTaskScheduler();
+        if (executorService != null) {
```
sdkottegoda: Can this be null?

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/util/MicroIntegratorBaseUtils.java
Nirothipan: yes , since the synapse environment is initialized finally and this is in the core. Also we call this method only once and the null check won't be costly.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/core/internal/TasksDSComponent.java
Nirothipan: yes if coordination is not enabled

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/TaskQueryHelper.java
sdkottegoda: Don't we need to capture that integrity violation and may be have a debug log and process each one? I mean this should be called when each task is deployed, isn't it? So does the batch update really contain a batch?

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/util/MicroIntegratorBaseUtils.java
sdkottegoda: but can the method 'getSynapseEnvironment' be called before the synapseEnvironment comes? It's not about the cost, this will hide erroneous situations where the synapseEnvironment was not set correctly.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/TaskQueryHelper.java
Nirothipan: we can figure out from result set whether it is udpated or not. We will need batch updates when adding valid tasks which were missed during db connection lost time.

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/util/MicroIntegratorBaseUtils.java
Nirothipan: In coordination code the flow is as follows, 

Register Task Service -> Register Synapse Environment -> Deploy Tasks -> Then it will call this method to get Synapse Environment. ( So null issue won't be there ).

But since this method is in BaseUtils -> If someone uses it not properly there will be NPE. 
I preferred returning null instead of running into NPE in runtime. wdyt . Else we can add a comment saying that this will return NPE if called before initializing synapse environment. wdyt

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/TaskQueryHelper.java
sdkottegoda: which were missed in the sense, deployed while the connection was lost?

Nirothipan: yes

sdkottegoda: This a 1% situation right? I feel like it's more risky to ignore than it is costly to catch it and log it. Basically IGNORE doesn't only ignore integrity violations it also ignores other types so I don't want it to be hidden. It's your call, 2 sides of the coin.

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/util/MicroIntegratorBaseUtils.java
sdkottegoda: We need to have NPE if someone uses it not properly because that is a design failure. Otherwise this will return null, and we might have done null checks somewhere else at it will fail at a lower stage without giving us no clue about the synapseEnvironment being null.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/db/TaskQueryHelper.java
Nirothipan: Fixed in https://github.com/wso2/micro-integrator/pull/1227/commits/922fa59ce6aa0e6b2ba7c9f1e90041a79628e2be

Issue #1169, opened 2020-02-24
Fix Integration Test Failures in windows


Nirothipan:
$subject

PR #1244, opened 2020-03-16
Fix JMSTransportProxyTestCase failure

sajinieKavindya:
## Purpose
$subject.

Related issue #1169


sdkottegoda:
Shall we add the tests cases that are fixed, in the description?

PR #1256, opened 2020-03-18
Fix proper shutdown of the server in integration tests

sajinieKavindya:
## Purpose
There can be occasions where the server does not get shut down properly even after waiting for 180s after executing stop command once. At that point, the server prints the log as _"Server stopped successfully ..."_ even though it is not properly shut down.

This PR addresses the above issue by trying to execute the stop command multiple times until it gets properly shutdown.  Here, assuming the server takes around 5s to stop the server, we are waiting for 180s in total until the server gets shut down. 

If the server still not gets shut down properly, we will throw a RuntimeException with the message _"Failed shutting down the sever"_.

### DIFF
--- a/integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
+++ b/integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
@@ -238,9 +238,20 @@ public class CarbonServerManager {
             log.info("Shutting down server ...");
 
             try {
+                // Waiting for 3 mins max until the server gets shut down
+                int retryCount = 36;
+                for (int i = 0; i < retryCount; i++) {
+                    if (isRemotePortInUse("localhost", managementPort)) {
+                        startProcess(carbonHome, getStartScriptCommand("stop"));
+                        waitTill(() -> isRemotePortInUse("localhost", managementPort), 5, TimeUnit.SECONDS);
+                    } else {
+                        break;
+                    }
+                }
 
-                startProcess(carbonHome, getStartScriptCommand("stop"));
-                waitTill(() -> isRemotePortInUse("localhost", managementPort), 180, TimeUnit.SECONDS);
+                if (isRemotePortInUse("localhost", managementPort)) {
+                    throw new AutomationFrameworkException("Failed shutting down the sever");
+                }
 
                 log.info("Server stopped successfully ...");
 


integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
```diff
@@ -238,9 +238,21 @@ public synchronized void serverShutdown(int portOffset, boolean skipRunningCodeC
             log.info("Shutting down server ...");
 
             try {
+                /* Trying to stop the server for multiple times until it gets properly shutdown. Assuming server takes
+                around 5s to stop the server, waiting for 180s in total until the server gets shut down */
+                int retryCount = 36;
+                for (int i = 0; i < retryCount; i++) {
+                    if (isRemotePortInUse("localhost", managementPort)) {
```
sdkottegoda: Shouldn't this be !isRemotePortInUse?

integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
```diff
@@ -238,9 +238,21 @@ public synchronized void serverShutdown(int portOffset, boolean skipRunningCodeC
...
+                        startProcess(carbonHome, getStartScriptCommand("stop"));
+                        waitTill(() -> isRemotePortInUse("localhost", managementPort), 5, TimeUnit.SECONDS);
```
sdkottegoda: shouldn't we wait till !isRemotePortInUse?

integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
```diff
@@ -238,9 +238,21 @@ public synchronized void serverShutdown(int portOffset, boolean skipRunningCodeC
             log.info("Shutting down server ...");
 
             try {
+                /* Trying to stop the server for multiple times until it gets properly shutdown. Assuming server takes
+                around 5s to stop the server, waiting for 180s in total until the server gets shut down */
```
Nirothipan: moving forward this comment won't be use ful. Shall we remove it or say we wait for 3 mins.

integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
sajinieKavindya: if the port is still in use is the server has not shut down yet, isn't it?

integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
sajinieKavindya: Changed the comment.

integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
sajinieKavindya: so basically what waitTill() method does is it does a TimeUnit.SECONDS.sleep(1) until the first predicate which is isRemotePortInUse("localhost", managementPort) is true. Once it is false, waitTill() method will stop waiting and continue.

Issue #1258, opened 2020-03-18
Template internal-api.xml


Nirothipan:
**Description:**
$subject so that it can be configured via deployment toml file.

PR #1262, opened 2020-03-19
Template internal api xml

Nirothipan:
Fixes https://github.com/wso2/micro-integrator/issues/1258


distribution/src/resources/config-tool/templates/conf/internal-apis.xml.j2
```diff
@@ -0,0 +1,99 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+<!--
+    This is the configuration file for internal APIs
+-->
+<internalApis>
+    <apis>
+    {% for api in internal_api %}
```
madushadhanushka: indentation issues

Issue #1183, opened 2020-02-27
Environment variable support for configuration files 


Vathsan:
**Description:**
Currently, following files support parameterize  configurations with system properties:
1. axis2.xml
2. carbon.xml
3. master-datasource.xml

example: 
```xml
<HostName>${system_server_hostname}</HostName>
```
To make EI more container friendly we **must** support to parameterize over environment variables as well.
Precedence should be as follows: Environment variable, System property, Hardcoded value
 
This is small fix to support above listed files. 
Ultimately we need to support this for all the configurations files that might be changed per environment.

Following files are must:
1. axis2.xml
2. carbon.xml
3. master-datasource.xml (files in datasource directory)
4. user-mgt.xml
5. jndi.properties

Optionally:
1. carbon.properties (optional)

We have already done the improvement for EI [1] 

[1] https://github.com/wso2/product-ei/issues/4724

PR #1265, opened 2020-03-19
Add environment variables support for configuration files

Vathsan:
Related issue https://github.com/wso2/micro-integrator/issues/1183


components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/core/util/CarbonUtils.java
```diff
@@ -113,12 +120,40 @@ public static void resolveLeafNodeValue(Node node) {
     }
 
     public static String resolveSystemProperty(String text) {
+        String sysRefs = StringUtils.substringBetween(text, SYS_PROPERTY_PLACEHOLDER_PREFIX, PLACEHOLDER_SUFFIX);
```
Nirothipan: Don't we need to validate pattern here

Issue #1253, opened 2020-03-17
Certificate issue while trying to log in to the MI Monitoring Dashboard


Vathsan:
**Description:**
When trying to login to MI Monitoring Dashboard with a self-signed certificate the following error occurs,

![image](https://user-images.githubusercontent.com/29911181/76856367-20de6a00-6879-11ea-830d-6b5eb75500fb.png)

We could observe the following logs in the dashboard console, 

```
ERROR {org.wso2.transport.http.netty.contractimpl.listener.WebSocketServerHandshakeHandler} - Error during WebSocket server handshake io.netty.handler.codec.DecoderException: javax.net.ssl.SSLException: Received fatal alert: bad_certificate
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:459)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:265)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1359)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:935)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:141)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:645)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:580)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:497)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:459)
[...]

PR #1268, opened 2020-03-20
Update carbon kernel version

Vathsan:
Fixes https://github.com/wso2/micro-integrator/issues/1253


Nirothipan:
AFIR  have the certs hard coded in some integrations tests . they might fail . Also without replacing them , we can copy from kernel in pom.

PR #1270, opened 2020-03-20
Make triggering unresponsive event immediate

Nirothipan:
## Purpose

$subject.

Coordination component wait for maximum heart beat retry time  ( default 15 seconds ) to trigger the member unresponsive event.

However, when considering scheduled tasks we need to trigger it as soon as possible so that we stop all running tasks immediately. 

Hence this PR is to make it spontaneous.

### DIFF
--- a/components/org.wso2.micro.integrator.coordination/src/main/java/org/wso2/micro/integrator/coordination/RDBMSMemberEventListenerTask.java
+++ b/components/org.wso2.micro.integrator.coordination/src/main/java/org/wso2/micro/integrator/coordination/RDBMSMemberEventListenerTask.java
@@ -47,11 +47,6 @@ public class RDBMSMemberEventListenerTask implements Runnable {
     private String localGroupId;
 
     /**
-     * Heart bear max retry time of the node
-     */
-    private int heartbeatMaxRetryTime;
-
-    /**
      * Communication bus object to communicate with the database for the context store.
      */
     private RDBMSCommunicationBusContextImpl communicationBusContext;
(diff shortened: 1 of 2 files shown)


components/org.wso2.micro.integrator.coordination/src/main/java/org/wso2/micro/integrator/coordination/RDBMSMemberEventListenerTask.java
```diff
@@ -106,19 +98,12 @@ public RDBMSMemberEventListenerTask(String nodeId, String localGroupId, int hear
...
-            log.warn("Error occurred while reading membership events. ", e);
-            if (inactiveTimestamp.equals(0L)) {
-                inactiveTimestamp = System.currentTimeMillis();
-            }
-            long inactiveHeartbeatAge = System.currentTimeMillis() - inactiveTimestamp;
-            if ((inactiveHeartbeatAge > heartbeatMaxRetryTime) && !wasMemberUnresponsive) {
-                log.warn("Node became unresponsive due to not being able to read events from database");
-                inactiveTimestamp = 0L; // reset in active time stamp
-                wasMemberUnresponsive = true;
+            log.error("Error occurred while reading membership events. ", e);
```
sdkottegoda: Should we do this for any throwable? shouldn't we only do this for and error during, readMembershipEvents()? And print an ERROR/WARN for exceptions during notifyWhatever event?

Nirothipan: readMembershipEvents throws a run time exception so we can't catch it and do this. Also we have no point in continuing when exception happens in readMembershipEvents.

PR #1279, opened 2020-03-21
Update connected nodes to skip itself when printing details

Nirothipan:


sdkottegoda:
Why are we doing this?

Nirothipan:
> Why are we doing this?

All the nodes will get itself printed ass member added when they joined the cluster, we get the following logs when they join cluster and this is unnecessarry.

[2020-03-23 10:34:03,932]  INFO {RDBMSCoordinationStrategy} - Successfully joined the cluster with id [149fa588-6aa2-4ac9-aa74-b6b1944577aa]

Also for the connected node details, we don't need to print that this node connected with itself upon server start which is meaningless after above log.

PR #1280, opened 2020-03-21
Update default heartbeat interval and retry count

Nirothipan:
Since 5000 is too much when considering task coordination as member removal will be triggered after heartbeat interval * retry count. Hence reducing it. This is still configurable.

### DIFF
--- a/components/org.wso2.micro.integrator.coordination/src/main/java/org/wso2/micro/integrator/coordination/util/RDBMSConstantUtils.java
+++ b/components/org.wso2.micro.integrator.coordination/src/main/java/org/wso2/micro/integrator/coordination/util/RDBMSConstantUtils.java
@@ -47,8 +47,8 @@ public class RDBMSConstantUtils {
     public static final String LOCAL_GROUP_ID = "localGroupId";
     public static final String SCHEDULED_PERIOD = "scheduledPeriod";
 
-    public static final int DEFAULT_HEART_BEAT_INTERVAL = 5000;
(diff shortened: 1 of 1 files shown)


sdkottegoda:
Are we reversing https://github.com/wso2/micro-integrator/pull/1270 here? Why? What is the problem of having a gap between unresponsive and member left events?

Nirothipan:
> Are we reversing #1270 here? Why? What is the problem of having a gap between unresponsive and member left events?

yes. Since the time interval is considerable ( 15 sec ) by default , if there are some crons between that it might get missed.

components/org.wso2.micro.integrator.coordination/src/main/java/org/wso2/micro/integrator/coordination/util/RDBMSConstantUtils.java
```diff
@@ -47,8 +47,8 @@
     public static final String LOCAL_GROUP_ID = "localGroupId";
     public static final String SCHEDULED_PERIOD = "scheduledPeriod";
 
-    public static final int DEFAULT_HEART_BEAT_INTERVAL = 5000;
-    public static final int DEFAULT_HEART_BEAT_MAX_RETRY = 3;
+    public static final int DEFAULT_HEART_BEAT_INTERVAL = 2000;
```
sdkottegoda: We need to test this thoroughly, This means that the heartbeat operation cannot exceed 2 seconds.

PR #1304, opened 2020-03-24
Handle task pause and resume in message processors for coordinated tasks.

Nirothipan:
The flow will be as follows,

- When deactivation is triggered from outside ( calling an api ) , if it is  a coordinated task the state of the task will be updated as DEACTIVATED in task store.

- The Node which is running the task, will pick up the tasks in DEACTIVATED states and will pause it and update the state as PAUSED.

- When activation is triggered from outside ( calling an api ) , if it is  a coordinated task the state of the task will be updated as ACTIVATED in task store.

- The Node which is assigned with the task, will pick up the tasks in ACTIVATED states and will schedule it and update the state as RUNNING.

### DIFF
--- a/components/mediation/tasks/org.wso2.micro.integrator.mediation.ntask/src/main/java/org/wso2/micro/integrator/mediation/ntask/NTaskTaskManager.java
+++ b/components/mediation/tasks/org.wso2.micro.integrator.mediation.ntask/src/main/java/org/wso2/micro/integrator/mediation/ntask/NTaskTaskManager.java
@@ -45,25 +45,19 @@ public class NTaskTaskManager implements TaskManager, TaskServiceObserver, Serve
 
     private static final Log logger = LogFactory.getLog(NTaskTaskManager.class.getName());
 
-    /**
-     * TODO ClusterGroupCommunicator NTASK_P2P_COMM_EXECUTOR is private in the carbon-commons-4.4.1 release, 
-     * this should be changed to use ClusterGroupCommunicator.NTASK_P2P_COMM_EXECUTOR.
-     */
-    private static final String NTASK_P2P_COMM_EXECUTOR = "__NTASK_P2P_COMM_EXECUTOR__";
-
     private String name;
 
     private boolean initialized = false;
 
     private org.wso2.micro.integrator.ntask.core.TaskManager taskManager;
 
-    private final Map<String, Object> properties = new HashMap<String, Object>(5);
+    private final Map<String, Object> properties = new HashMap<>(5);
 
-    protected final Properties configProperties = new Properties();
+    private final Properties configProperties = new Properties();
 
-    private final List<TaskManagerObserver> observers = new ArrayList<TaskManagerObserver>();
+    private final List<TaskManagerObserver> observers = new ArrayList<>();
 
-    private final List<TaskDescription> taskQueue = new ArrayList<TaskDescription>();
+    private final List<TaskDescription> taskQueue = new ArrayList<>();
 
     private final Object taskQueueLock = new Object();
 
@@ -205,7 +199,7 @@ public class NTaskTaskManager implements TaskManager, TaskServiceObserver, Serve
                     logger.warn("#pause Could not pause task [" + taskName + "]. Task manager is not available.");
                     return false;
                 }
-                taskManager.pauseTask(taskName);
+                taskManager.handleTaskPause(taskName);
             }
             return true;
         } catch (Exception e) {
@@ -227,7 +221,8 @@ public class NTaskTaskManager implements TaskManager, TaskServiceObserver, Serve
                 }
                 List<TaskInfo> taskList = taskManager.getAllTasks();
                 for (TaskInfo taskInfo : taskList) {
-                    taskManager.pauseTask(taskInfo.getName());
+                    String taskName = taskInfo.getName();
+                    taskManager.handleTaskPause(taskName);
                 }
             }
             return true;
@@ -251,7 +246,7 @@ public class NTaskTaskManager implements TaskManager, TaskServiceObserver, Serve
                     logger.warn("#resume Could not resume task [" + taskName + "]. Task manager is not available.");
                     return false;
                 }
-                taskManager.resumeTask(taskName);
+                taskManager.handleTaskResume(taskName);
             }
         } catch (Exception e) {
             logger.error("Cannot resume task [" + taskName + "]. Error: " + e.getLocalizedMessage(), e);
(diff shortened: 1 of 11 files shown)


components/mediation/tasks/org.wso2.micro.integrator.mediation.ntask/src/main/java/org/wso2/micro/integrator/mediation/ntask/NTaskTaskManager.java
```diff
@@ -227,7 +223,10 @@ public boolean pauseAll() {
                 }
                 List<TaskInfo> taskList = taskManager.getAllTasks();
                 for (TaskInfo taskInfo : taskList) {
-                    taskManager.pauseTask(taskInfo.getName());
+                    String taskName = taskInfo.getName();
+                    if (!taskManager.isDeactivated(taskName)) {
```
sdkottegoda: Do we need this validation here, it seems like a validation that needs to happen inside the task manager

components/mediation/tasks/org.wso2.micro.integrator.mediation.ntask/src/main/java/org/wso2/micro/integrator/mediation/ntask/NTaskTaskManager.java
```diff
@@ -574,8 +573,7 @@ public boolean isTaskDeactivated(String taskName) {
                 return false;
             }
             try {
-                return taskManager.getTaskState(taskName)
-                                  .equals(org.wso2.micro.integrator.ntask.core.TaskManager.TaskState.PAUSED);
+                return taskManager.isDeactivated(taskName);
```
sdkottegoda: shouldn't this be isPaused()

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/scehduler/CoordinatedTaskScheduler.java
```diff
@@ -89,12 +94,44 @@ public void run() {
...
+        ScheduledTaskManager scheduledTaskManager = taskManager;
+        List<String> pausedTasks = new ArrayList<>();
+        List<String> deployedTasks = taskManager.getAllCoordinatedTasksDeployed();
+        deactivatedTasks.forEach(task -> {
+            if (deployedTasks.contains(task)) {
+                try {
+                    scheduledTaskManager.stopExecution(task);
+                    pausedTasks.add(task);
+                } catch (TaskException e) {
+                    LOG.error("Error stopping the task [" + task + "]", e);
```
sdkottegoda: If we get an error here, the exception will be logged and ignored, but the state in the DB will get updated to PAUSED, is that ok?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/scehduler/CoordinatedTaskScheduler.java
```diff
@@ -89,12 +94,44 @@ public void run() {
...
+                }
+            } else {
+                LOG.info("The task [" + task + "] retrieved to be paused is not a deployed task "
```
sdkottegoda: same here... Shouldn't we do something about this state?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/core/TaskManager.java
```diff
@@ -30,41 +30,41 @@
...
      */
-    public boolean deleteTask(String taskName) throws TaskException;
+    boolean deleteTask(String taskName) throws TaskException;
 
     /**
      * Handles the task with given name. Schedule if its not coordinated else update the task DB.
      *
      * @param taskName The name of the task
-     * @throws TaskException - Exception
+     * @throws TaskException - Exception Exception
```
sdkottegoda: Exception Exception

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/core/TaskManager.java
```diff
@@ -30,41 +30,41 @@
     /**
      * Initialize the startup tasks.
      *
-     * @throws TaskException
+     * @throws TaskException - Exception
```
sdkottegoda: Standard notation doesn't have '-' does it?

components/mediation/tasks/org.wso2.micro.integrator.mediation.ntask/src/main/java/org/wso2/micro/integrator/mediation/ntask/NTaskTaskManager.java
Nirothipan: moved to task manager .

components/mediation/tasks/org.wso2.micro.integrator.mediation.ntask/src/main/java/org/wso2/micro/integrator/mediation/ntask/NTaskTaskManager.java
Nirothipan: This method checks whether the task is deactivated or not. For standalone tasks it is whether only paused for others we have to check other conditions as well. Hence I named it as isDeactivated.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/scehduler/CoordinatedTaskScheduler.java
Nirothipan: It will not get updated, since we are adding the list of tasks which are paused successfully to the list **pausedTasks.add(task);** and then only updating them to db in line 132. If it is an exception, it won't get added to the list.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/scehduler/CoordinatedTaskScheduler.java
Nirothipan: Not really, if it hits this else condition , for some reason the task retrieved from store is not in the deployed list. It could be due to deployment delay or an invalid entry in the store. So we just print info log and ignore , so that task store will clean it out.

PR #1307, opened 2020-03-25
Fix transport test failures in Windows due to server initialization errors

sdkottegoda:
## Purpose
The server startup hangs intermittently in windows without a specific reason. The fix is to retry server startup for 3 attempts and fail if it still doesn't start. This fixes the test failures due to server initialization failure in windows.


integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
```diff
@@ -148,7 +150,17 @@ public synchronized void startServerUsingCarbonHome(String carbonHome, Map<Strin
             waitTill(() -> !isRemotePortInUse("localhost", managementPort), 180, TimeUnit.SECONDS);
 
             if (!isRemotePortInUse("localhost", managementPort)) {
-                throw new RuntimeException("Server initialization failed");
+                if (retryAttempt < retryLimit){
+                    retryAttempt++;
+                    serverShutdown(portOffset, true);
+                    startServerUsingCarbonHome(carbonHome, commandMap);
```
Nirothipan: Shall we add some info logs when we retry

integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
```diff
@@ -148,7 +150,17 @@ public synchronized void startServerUsingCarbonHome(String carbonHome, Map<Strin
...
+                }
+                else {
+                    throw new RuntimeException("Server initialization failed");
```
Nirothipan: do we need to set retryAttempt to zero here also ?

integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
```diff
@@ -148,7 +150,17 @@ public synchronized void startServerUsingCarbonHome(String carbonHome, Map<Strin
...
+                }
+            }
+            else{
```
Nirothipan: Shall we format this code

integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
sdkottegoda: done

integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
sdkottegoda: done

PR #1313, opened 2020-03-26
Check whether results available before processing

Nirothipan:
$subject


components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/store/connector/RDMBSConnector.java
```diff
@@ -273,8 +273,10 @@ public void deleteTasks(List<String> tasks) throws TaskCoordinationException {
                 RETRIEVE_TASK_STATE)) {
             preparedStatement.setString(1, taskName);
             try (ResultSet resultSet = preparedStatement.executeQuery()) {
-                resultSet.next();
-                return CoordinatedTask.States.valueOf(resultSet.getString(TASK_STATE));
+                if (resultSet.next()) {
+                    return CoordinatedTask.States.valueOf(resultSet.getString(TASK_STATE));
+                }
+                return null;
```
abeykoon: Is returning `null` a valid case here? Or should we throw a TaskCoordinationException?

Nirothipan: There can be cases where the task is not present, so returning null. Its up to the caller to handle the state null.

PR #1330, opened 2020-03-30
Disable graceful Shutdown in MI integration tests

arunans23:
Graceful Shutdown feature will be added to MI with the following PR in synapse. https://github.com/wso2/wso2-synapse/pull/1478
We need to disable this feature in Integration Tests, since it will take longer time to complete if this is enabled.

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/integration/mediation-tests/tests-mediator-1/src/test/resources/automation.xml
+++ b/integration/mediation-tests/tests-mediator-1/src/test/resources/automation.xml
@@ -274,6 +274,7 @@
                     <parameter name="-DenableManagementApi" value="true"/>
 
                     <parameter name="startupScript" value="micro-integrator"/>
+                    <parameter name="-DgracefulShutdown" value="false"></parameter>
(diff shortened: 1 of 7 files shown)


Nirothipan:
Do we need to add this for DSS tests as well ?
Also does graceful shutdown works when we stop a background process ( like micro-integrator.sh stop ) as we are doing that in integration tests, just asking for knowledge ?

arunans23:
> Do we need to add this for DSS tests as well ?

We do not need this for DSS tests.

> Also does graceful shutdown works when we stop a background process ( like micro-integrator.sh stop ) as we are doing that in integration tests, just asking for knowledge ?

Yes, it works. When you do `micro-integrator stop` it sends a shutdown signal to the server. Hence it is a graceful shutdown. It does not kill the process instantly.

PR #1339, opened 2020-03-31
Update message processor activate and deactivate flow in task

Nirothipan:
**Improvements**

1.  When scheduling tasks, validate db update and react accordingly ( whether failed or not ).
2. When pausing tasks , if they are running locally just pause them and update DB as paused rather than doing all the steps in https://github.com/wso2/micro-integrator/pull/1304
3. While activating and deactivating, if the state is as running or paused respectively ignoring that update so that the relavant nodes doesn't have to retrieve them again and act accordingly.

### DIFF
--- a/components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/scehduler/CoordinatedTaskScheduler.java
+++ b/components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/scehduler/CoordinatedTaskScheduler.java
@@ -32,7 +32,6 @@ import org.wso2.micro.integrator.ntask.core.impl.standalone.ScheduledTaskManager
 import org.wso2.micro.integrator.ntask.core.internal.DataHolder;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -173,6 +172,7 @@ public class CoordinatedTaskScheduler implements Runnable {
             return;
         }
         List<String> deployedCoordinatedTasks = taskManager.getAllCoordinatedTasksDeployed();
+        List<String> erroredTasks = new ArrayList<>();
         for (String taskName : tasksOfThisNode) {
             if (deployedCoordinatedTasks.contains(taskName)) {
                 if (LOG.isDebugEnabled()) {
@@ -182,7 +182,7 @@ public class CoordinatedTaskScheduler implements Runnable {
                     taskManager.scheduleCoordinatedTask(taskName);
                 } catch (TaskException ex) {
                     if (!TaskException.Code.DATABASE_ERROR.equals(ex.getCode())) {
-                        taskStore.updateTaskState(Collections.singletonList(taskName), CoordinatedTask.States.NONE);
+                        erroredTasks.add(taskName);
                     }
                     LOG.error("Exception occurred while scheduling coordinated task : " + taskName, ex);
                 }
@@ -191,6 +191,7 @@ public class CoordinatedTaskScheduler implements Runnable {
                                  + "in this node or an invalid entry, hence ignoring it.");
             }
         }
+        taskStore.updateTaskState(erroredTasks, CoordinatedTask.States.NONE);
     }
 
     /**
--- a/components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/scehduler/CoordinatedTaskScheduler.java
+++ b/components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/scehduler/CoordinatedTaskScheduler.java
@@ -32,6 +32,7 @@ import org.wso2.micro.integrator.ntask.core.impl.standalone.ScheduledTaskManager
 import org.wso2.micro.integrator.ntask.core.internal.DataHolder;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/scehduler/CoordinatedTaskScheduler.java
```diff
@@ -172,21 +173,24 @@ private void scheduleAssignedTasks(CoordinatedTask.States state) throws TaskCoor
...
                 if (LOG.isDebugEnabled()) {
                     LOG.debug("Submitting retrieved task [" + taskName + "] to the task manager.");
                 }
                 try {
                     taskManager.scheduleCoordinatedTask(taskName);
-                } catch (TaskException e) {
-                    LOG.error("Exception occurred while scheduling coordinated task : " + taskName, e);
+                } catch (TaskException ex) {
+                    if (!TaskException.Code.DATABASE_ERROR.equals(ex.getCode())) {
+                        taskStore.updateTaskState(Collections.singletonList(taskName), CoordinatedTask.States.NONE);
```
sdkottegoda, line 185: What if we add these to a failed list and perform this at the end of the loop, will reduce the time for the db callse? Also, don't we need to unassign the task in this case?

sdkottegoda: Or do we need a logic like, if scheduling a task fails 2/3 times consecutively, we will unassign it?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/store/connector/RDMBSConnector.java
```diff
@@ -181,6 +183,36 @@ public void deleteTasks(String nodeId) throws TaskCoordinationException {
...
     }
 
+    /**
+     * Deactivates the task.
+     *
+     * @param name - Name of the task.
+     */
+    public void deactivateTask(String name) throws TaskCoordinationException {
+        try (Connection connection = getConnection(); PreparedStatement preparedStatement = connection.prepareStatement(
+                DEACTIVATE_TASK)) {
```
sdkottegoda, line 193: I prefer the name UPDATE_TASK_STATUS_TO_DEACTIVATED since it describes to query better. WDYT?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/store/connector/RDMBSConnector.java
```diff
@@ -351,6 +360,34 @@ public void updateAssignmentAndState(Map<String, String> tasks) throws TaskCoord
...
+    /**
+     * Update the state of task.
+     *
+     * @param taskName     Name of the task.
+     * @param updatedState Updated state.
+     * @param destinedId   Destined Node Id.
+     * @return True if update is successful.
+     * @throws TaskCoordinationException when something goes wrong while updating.
+     */
+    public boolean updateTaskState(String taskName, CoordinatedTask.States updatedState, String destinedId)
```
sdkottegoda, line 372: Why are we passing both the name and the destinedId? In case if the assignment has changed in the meantime?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/core/impl/AbstractQuartzTaskManager.java
```diff
@@ -282,7 +282,7 @@ protected synchronized void resumeLocalTask(String taskName) throws TaskExceptio
         }
     }
 
-    private boolean containsLocalTask(String taskName, String taskGroup) throws TaskException {
+    protected boolean isPreviouslyScheduled(String taskName, String taskGroup) throws TaskException {
```
sdkottegoda, line 285: what does previosulyScheduled mean?

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/store/connector/RDMBSConnector.java
Nirothipan: yes , to handle some edge cases. If everything is idea, assignment shouldn't have changed. If it is changed, we shouldn't be running it locally.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/core/impl/AbstractQuartzTaskManager.java
Nirothipan: The task was already scheduled and ran in this node and has been stopped for some reason.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/coordination/task/scehduler/CoordinatedTaskScheduler.java
Nirothipan: +1 will move to end of loop. No point in un-assigning the task , If the node leaves the cluster, it will get assigned to something else automatically . If not , even if we un assign  it might get assigned to the same node again.

components/mediation/tasks/org.wso2.micro.integrator.ntask.core/src/main/java/org/wso2/micro/integrator/ntask/core/impl/AbstractQuartzTaskManager.java
sdkottegoda: previouslyScheduledButFailed then? I didn't understand it previously, that's why

Nirothipan: its not failed , but has been stopped . The method name isPreviouslyScheduled is precise, it gives the result whether the task has been scheduled already in this node already or not.

Issue #1098, opened 2019-12-27
Support for Kubernetes vault secrets implementations


abeykoon:
**Description:**

Rather using WSO2 way to secure the passwords (and keep the master password in a plain text file as we need to start the container automatically :P), people would like to use 

* [Docker secrets](https://docs.docker.com/engine/swarm/secrets/) if on docker    
* Run a sidecar  with vault secrets of on K8S. [Hashicorp](https://www.hashicorp.com/blog/whats-next-for-vault-and-kubernetes/) is a popular choice. 

so that all their passwords are centrally managed. Thus it will be a good improvement if WSO2 MI can support those vault integrations Out of the Box. We can even consider developing a separate sideCar, so that anyone who need  vault integration can use. 

**Affected Product Version:**
MI 1.1.0

**OS, DB, other environment details and versions:**    
N/A

**Steps to reproduce:**
N/A

abeykoon:
Can we configure Hashicorp  as the vault in this approach? 
[https://kubernetes.io/docs/concepts/configuration/secret/](https://kubernetes.io/docs/concepts/configuration/secret/)
[https://www.hashicorp.com/blog/injecting-vault-secrets-into-kubernetes-pods-via-a-sidecar/](https://www.hashicorp.com/blog/injecting-vault-secrets-into-kubernetes-pods-via-a-sidecar/)

sachithKay:
No this provides support for kubernetes and docker secrets only. We are not providing out of the box support for hashicorp. We will provide an extension point for users to implement third party vaults. Closing this issue since we already provide support docker and k8 secret support.

PR #1361, opened 2020-04-06
Add env variable support for secret resolving

sachithKay:
Fixes: https://github.com/wso2/micro-integrator/issues/1098

```
wso2:vault-lookup('<alias>', '<type>', '<isEncripted>')
```
alias - environment variable name
type - ENV

### DIFF
--- a/components/mediation/security/org.wso2.micro.integrator.mediation.security/src/main/java/org/wso2/micro/integrator/mediation/security/vault/EnvironmentSecretRepository.java
+++ b/components/mediation/security/org.wso2.micro.integrator.mediation.security/src/main/java/org/wso2/micro/integrator/mediation/security/vault/EnvironmentSecretRepository.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.micro.integrator.mediation.security.vault;
+
+import org.wso2.securevault.secret.SecretRepository;
+import java.util.Properties;
+
+public class EnvironmentSecretRepository implements SecretRepository {
+
+    /* Parent secret repository */
+    private SecretRepository parentRepository;
+
+    @Override
+    public void init(Properties properties, String s) {
+        // Nothing to do here
+    }
+
+    @Override
+    public String getSecret(String alias) {
+        String rawValue = getPlainTextSecret(alias);
+        return SecureVaultUtils.decryptSecret(rawValue);
+    }
+
+    /**
+     * Returns the plaintext value from the environment.
+     *
+     * @param alias Environment variable name.
+     */
+    public String getPlainTextSecret(String alias) {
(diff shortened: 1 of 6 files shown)


components/mediation/security/org.wso2.micro.integrator.mediation.security/src/main/java/org/wso2/micro/integrator/mediation/security/vault/EnvironmentSecretRepository.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Returns the plaintext value from the environment.
+     *
+     * @param alias Environment variable name.
+     */
+    public String getPlainTextSecret(String alias) {
+        String planTextSecret = System.getenv(alias);
+        if (planTextSecret == null || planTextSecret.isEmpty()) {
+            throw new SynapseException("Plain text secret value has not been set for variable " + alias);
```
Nirothipan, line 61: Should we throw an exception here  or just return whatever comes as null or empty. The user might write the logic based on the availability of multiple env properties.

components/mediation/security/org.wso2.micro.integrator.mediation.security/src/main/java/org/wso2/micro/integrator/mediation/security/vault/EnvironmentSecretRepository.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public String getSecret(String alias) {
+        String rawValue = getPlainTextSecret(alias);
+        DecryptionProvider decyptProvider = CipherInitializer.getInstance().getDecryptionProvider();
+        if (decyptProvider == null) {
+            // This cannot happen unless someone mess with OSGI references
+            LOG.error("Can not proceed decryption due to the secret repository initialization error");
+            throw new SynapseException("Secret repository is null for alias " + alias);
```
Nirothipan, line 48: alias doesn't make any impact here , isn't it . We can just say it it null.
However do we need this null check , if there are some osgi issues it will be caught in the lower level and we don't need to check null for each run ? wdyt

components/mediation/security/org.wso2.micro.integrator.mediation.security/src/main/java/org/wso2/micro/integrator/mediation/security/vault/EnvironmentSecretRepository.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+        return new String(decyptProvider.decrypt(rawValue.trim().getBytes()));
```
Nirothipan, line 50: Shall we extract this code to a common method https://github.com/wso2/micro-integrator/blob/master/components/mediation/security/org.wso2.micro.integrator.mediation.security/src/main/java/org/wso2/micro/integrator/mediation/security/vault/FileSecretRepository.java#L61.L70

components/mediation/security/org.wso2.micro.integrator.mediation.security/src/main/java/org/wso2/micro/integrator/mediation/security/vault/EnvironmentSecretRepository.java
sachithKay: We can remove the exception message. But I think its better if we keep the null check.

components/mediation/security/org.wso2.micro.integrator.mediation.security/src/main/java/org/wso2/micro/integrator/mediation/security/vault/EnvironmentSecretRepository.java
sachithKay: +1. done

components/mediation/security/org.wso2.micro.integrator.mediation.security/src/main/java/org/wso2/micro/integrator/mediation/security/vault/EnvironmentSecretRepository.java
sachithKay: done

PR #1362, opened 2020-04-06
Add tests for coordinated tasks

Nirothipan:
### Following set of test cases are included.

- Test whether a task is scheduled only in a single node.
- Test whether all the tasks are scheduled in single node ( Active passive resolver ).
- Test whether the task gets executed in the scheduled node.
- Test whether task count is honored in execution.
- Test whether the task gets rescheduled when the executing node leaves the cluster.
- Test whether the completed task does not get rescheduled when the executing node leaves the cluster.
- Test whether a pinned task doesn’t get rescheduled when the pinned server leaves the cluster.

### Updates for test framework 
- Start servers with deployment synchronization and registry sharing when more nodes added to the cluster.


integration/mediation-tests/tests-platform/tests-coordination/src/test/java/org/wso2/micro/integrator/tasks/TaskTests.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static org.wso2.micro.integrator.TestUtils.CLUSTER_DEP_TIMEOUT;
+import static org.wso2.micro.integrator.TestUtils.CLUSTER_TASK_RESCHEDULE_TIMEOUT;
+import static org.wso2.micro.integrator.TestUtils.isTaskExistInStore;
+import static org.wso2.micro.integrator.TestUtils.deployTasks;
+import static org.wso2.micro.integrator.TestUtils.deploymentLog;
+import static org.wso2.micro.integrator.TestUtils.getNode;
+
+public class TaskTests extends ESBIntegrationTest {
```
abeykoon, line 42: Better add a class level description what these tests are about

PR #1389, opened 2020-04-11
Define poi import version to be the same as dependency version

sdkottegoda:
## Purpose
Define poi import version to be the same as dependency version which is 3.17.0-wso2v1


Nirothipan:
@sdkottegoda seems this is closed when you did a force push .

Issue #1377, opened 2020-04-09
NoClassDef Error while deploying MongoDB DSS


madushadhanushka:
**Description:**
The following NoClassDefFound error occurred while deploying DSS with MongoDB.
```
ERROR {org.wso2.micro.integrator.dataservices.core.DBDeployer} - The Mongodb-1.0.0.dbs service, which is not valid, caused {1} java.lang.NoClassDefFoundError: de/undercouch/bson4jackson/BsonFactory
	at org.jongo.Jongo.<init>(Jongo.java:33)
	at org.wso2.micro.integrator.dataservices.core.description.config.MongoConfig.<init>(MongoConfig.java:80)
	at org.wso2.micro.integrator.dataservices.core.description.config.ConfigFactory.getMongoConfig(ConfigFactory.java:103)
	at org.wso2.micro.integrator.dataservices.core.description.config.ConfigFactory.createConfig(ConfigFactory.java:63)
	at org.wso2.micro.integrator.dataservices.core.DataServiceFactory.createDataService(DataServiceFactory.java:159)
	at org.wso2.micro.integrator.dataservices.core.DBDeployer.createDBService(DBDeployer.java:806)
	at org.wso2.micro.integrator.dataservices.core.DBDeployer.processService(DBDeployer.java:1164)
	at org.wso2.micro.integrator.dataservices.core.DBDeployer.deploy(DBDeployer.java:196)
	at org.wso2.micro.application.deployer.handler.DefaultAppDeployer.deployRecursively(DefaultAppDeployer.java:174)
	at org.wso2.micro.application.deployer.handler.DefaultAppDeployer.deployArtifacts(DefaultAppDeployer.java:70)
	at org.wso2.micro.integrator.initializer.deployment.application.deployer.CAppDeploymentManager.deploy(CAppDeploymentManager.java:126)
	at org.wso2.micro.integrator.initializer.deployment.AppDeployerServiceComponent.activate(AppDeployerServiceComponent.java:80)
```

As a workaround, you can copy "de.undercouch.bson4jackson_2.9.0.jar" into the dropping folder.

**Steps to reproduce:**
Add following MongoDB dataservice dbs into the MI and try to deploy.
```
<data name="Mongodb" serviceNamespace="" serviceGroup="">
  <description />
  <config id="mongoldb">
    <property name="mongoDB_servers">localhost</property>
    <property name="mongoDB_database">mydb</property>
    <property name="mongoDB_write_concern">NONE</property>
[...]

PR #1413, opened 2020-04-16
Bundle de.undercouch.bson4jackson_2.9.0.jar

sajinieKavindya:
## Purpose
$subject.

Fixes https://github.com/wso2/micro-integrator/issues/1377

### DIFF
--- a/components/data/data-services/org.wso2.micro.integrator.dataservices.core/pom.xml
+++ b/components/data/data-services/org.wso2.micro.integrator.dataservices.core/pom.xml
@@ -233,6 +233,10 @@
             <groupId>org.apache.ws.security.wso2</groupId>
             <artifactId>wss4j</artifactId>
         </dependency>
+        <dependency>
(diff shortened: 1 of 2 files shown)


components/data/data-services/org.wso2.micro.integrator.dataservices.core/pom.xml
```diff
@@ -233,6 +233,24 @@
             <groupId>org.apache.ws.security.wso2</groupId>
             <artifactId>wss4j</artifactId>
         </dependency>
+        <dependency>
+            <groupId>de.undercouch</groupId>
+            <artifactId>bson4jackson</artifactId>
+            <exclusions>
```
Nirothipan, line 239: We don't need the exclusions here as we have in parent pom ?

sajinieKavindya: Removed the exclusions.

PR #1422, opened 2020-04-20
Create secure-vault.properties if not exists.

sachithKay:


sdkottegoda:
We need to update the documentation at https://ei.docs.wso2.com/en/next/micro-integrator/develop/creating-artifacts/encrypting-synapse-passwords/ too

Issue #1223, opened 2020-03-13
[ Coordination ] Make node id a configurable value


Nirothipan:
**Description:**
$subject.

Now this is random UUID .. we need to make it possible to be read from some configs and if not specified only generate a UUID.

 **this.localNodeId = generateRandomId();**

```java
private RDBMSCoordinationStrategy(RDBMSCommunicationBusContextImpl communicationBusContext) {
        readCoordinationConfigurations();
        if (heartbeatMaxRetry < 1) {
            throw new ClusterCoordinationException("heartbeatMaxRetry configuration should be larger than 0");
        }
        this.heartbeatMaxRetryInterval = heartBeatInterval * heartbeatMaxRetry;
        this.heartbeatWarningMargin = heartbeatMaxRetryInterval * 0.75;

        ThreadFactory namedThreadFactory = new ThreadFactoryBuilder().setPriority(7)
                                                                     .setNameFormat("RDBMSCoordinationStrategy-%d").build();
        this.threadExecutor = Executors.newSingleThreadExecutor(namedThreadFactory);

        this.localNodeId = generateRandomId();
        this.communicationBusContext = communicationBusContext;
        this.rdbmsMemberEventProcessor = new RDBMSMemberEventProcessor(localNodeId, localGroupId,
                                                                       heartbeatMaxRetryInterval, communicationBusContext);
    }
```

PR #1430, opened 2020-04-21
Make node id configurable

Nirothipan:
Fixes https://github.com/wso2/micro-integrator/issues/1223

## Purpose

Make the node id configurable.

**Toml config**
```toml
[cluster_config]
node_id = "node-1"
```

### DIFF
--- a/components/org.wso2.micro.integrator.coordination/src/main/java/org/wso2/micro/integrator/coordination/RDBMSCoordinationStrategy.java
+++ b/components/org.wso2.micro.integrator.coordination/src/main/java/org/wso2/micro/integrator/coordination/RDBMSCoordinationStrategy.java
@@ -21,6 +21,8 @@ package org.wso2.micro.integrator.coordination;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.wso2.config.mapper.ConfigParser;
+import org.wso2.micro.core.util.StringUtils;
 import org.wso2.micro.integrator.coordination.exception.ClusterCoordinationException;
 import org.wso2.micro.integrator.coordination.node.NodeDetail;
 import org.wso2.micro.integrator.coordination.util.MemberEventType;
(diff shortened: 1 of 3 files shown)


Vathsan:
Do we support configuring node id as following in toml file?

```
[cluster_config]
node_id = "$env{variable_name}"
```
and
```
[cluster_config]
node_id = "$sys{property_name}"
```

Nirothipan:
> Do we support configuring node id as following in toml file?
> 
> ```
> [cluster_config]
> node_id = "$env{variable_name}"
> ```
> 
> and
> 
> ```
> [cluster_config]
> node_id = "$sys{property_name}"
> ```

Yes and this is upto the parser to to resolve the sys or env and provide the values to the runtime as it does for other variables and this has nothing to do here. Here we are reading from the toml only.

Vathsan:
> > Do we support configuring node id as following in toml file?
> > ```
> > [cluster_config]
> > node_id = "$env{variable_name}"
> > ```
> > 
> > 
> > and
> > ```
> > [cluster_config]
> > node_id = "$sys{property_name}"
> > ```
> 
> Yes and this is upto the parser to to resolve the sys or env and provide the values to the runtime as it does for other variables and this has nothing to do here. Here we are reading from the toml only.

Yes, by default the parser resolves the above two configurations. As per the fix in [1] we can set `avoidResolvingEnvAndSysVariable` system property and avoid resolving environment variables and system properties.  In occasions where we pass the above system properties, we need to handle resolving node_id from the code level. WDYT?
[1] https://github.com/wso2/config-mapper/pull/13
[2] https://github.com/wso2/config-mapper/pull/14

Nirothipan:
> > > Do we support configuring node id as following in toml file?
> > > ```
> > > [cluster_config]
> > > node_id = "$env{variable_name}"
> > > ```
> > > 
> > > 
> > > and
> > > ```
> > > [cluster_config]
> > > node_id = "$sys{property_name}"
> > > ```
> > 
> > 
> > Yes and this is upto the parser to to resolve the sys or env and provide the values to the runtime as it does for other variables and this has nothing to do here. Here we are reading from the toml only.
> 
> Yes, by default the parser resolves the above two configurations. As per the fix in [1] we can set `avoidResolvingEnvAndSysVariable` system property and avoid resolving environment variables and system properties. In occasions where we pass the above system properties, we need to handle resolving node_id from the code level. WDYT?
> [1] [wso2/config-mapper#13](https://github.com/wso2/config-mapper/pull/13)

In the implementation we first check from system and then from env and then from toml ( using prop **nodeId** ). If the user has defined in toml as "$env" sthng has added the above mentioned property which is a contracting , then it is a user error. We can't cater that as we already have other options for them to defined node id as sys or env value.

PR #1439, opened 2020-04-22
Implement user management via Management API

sdkottegoda:
This PR bring the capability of managing users via the management api. The request/response formats are describe below.

# Add user
## Request
POST https://localhost:9164/management/users
{ 
   "userId":"user4",
   "password":"pwd1",
   "isAdmin":"true"
}

## Response
200: OK
{
    "userId":"user4",
    “status”:added
}
# Delete user
## Request
DELETE https://localhost:9164/management/users/{userid}
## Response
200: OK
{
    "userId":"user4",
    “status”:deleted
}
404: Not Found
# View Users
## Request
GET https://localhost:9164/management/users?pattern=”*”&role=admin
## Response
200: OK
{
	count:2 
	list:
	[	
		userId: user1,
		userId: user2,
		userId: user3,
	]
}
404: Not Found
# View Specific User
## Request
GET https://localhost:9164/management/users/{user1}
## Response
200: OK
{ 
userid: “user1”,
	isAdmin: true/false,
	roles : 
[
     role1,
    role2,
]
}
404: Not found


components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/ManagementApiParser.java
```diff
@@ -0,0 +1,148 @@
...
+    public static String getConfigurationFilePath() {
+        return new File(MicroIntegratorBaseUtils.getCarbonConfigDirPath(), "internal-apis.xml").getAbsolutePath();
+    }
+
+    public static OMElement getManagementApiElement() throws IOException, CarbonException, XMLStreamException,
+            ManagementApiUndefinedException {
+        if (Objects.nonNull(managementApiElement)) {
+            return managementApiElement;
+        }
+        LOG.debug("Parsing " + getConfigurationFilePath() + " for the first time.");
```
Nirothipan, line 72: We need to check whether debug is enabled here .

integration/mediation-tests/tests-platform/tests-user-management/README.txt
```diff
@@ -0,0 +1,6 @@
+1) Start a mysql server locally ( No need to create any data base ).
+2) Update the datasource details in
+<micro-integrator-repo>/integration/mediation-tests/tests-platform/tests-coordination/src/test/resources/artifacts/ESB/server/conf/deployment.toml
```
Nirothipan, line 3: tests-coordination ? : )

integration/mediation-tests/tests-platform/tests-user-management/README.txt
```diff
@@ -0,0 +1,6 @@
...
+file ( you may define any name for database and it will be created in the tests if it doesn't exist or will be
+dropped and created if it exists to make sure that we source the updated db script )
+3) Execute the command "mvn clean install -P cluster-tests"
```
Nirothipan, line 6: -P cluster-tests :octocat:

integration/mediation-tests/tests-platform/tests-user-management/src/test/resources/testng.xml
```diff
@@ -0,0 +1,38 @@
...
+        <listener class-name="org.wso2.carbon.automation.engine.testlisteners.TestManagerListener"/>
+        <listener class-name="org.wso2.carbon.automation.engine.testlisteners.TestReportListener"/>
+    </listeners>
+
+    <test name="Cluster-tests" verbose="2" parallel="none">
+        <classes>
+            <class name="org.wso2.micro.integrator.startup.StartupTests"/>
+            <class name="org.wso2.micro.integrator.tasks.TaskTests"/>
+            <class name="org.wso2.micro.integrator.message.processor.MessageProcessorTests"/>
+        </classes>
```
Nirothipan, line 35: Accident commits ?

integration/mediation-tests/tests-platform/tests-user-management/README.txt
sdkottegoda: Remove the whole integration folder

integration/mediation-tests/tests-platform/tests-user-management/src/test/resources/testng.xml
sdkottegoda: Removed

distribution/src/resources/config-tool/default.json
```diff
@@ -430,8 +430,8 @@
   "management_api.handler.token_store_config.remove_oldest_token_on_overflow": "true",
   "management_api.handler.token_config.expiry": "3600",
   "management_api.handler.token_config.size": "2048",
-  "management_api.handler.file_user_store.enable": true,
-  "management_api.handler.user_store.users": [
+  "management_api.file_user_store.enabled": true,
```
Nirothipan, line 433: Let's make this also as **enable** to be consistence with others ?

Nirothipan:
The test cases using management api would fail with authorization error. 
```
Caused by: java.io.IOException: Server returned HTTP response code: 401 for URL: https://localhost:9214/management/message-processors
```

distribution/src/dbscripts/mysql_user.sql
```diff
@@ -0,0 +1,253 @@
+CREATE TABLE UM_TENANT (
+			UM_ID INTEGER NOT NULL AUTO_INCREMENT,
+	        UM_DOMAIN_NAME VARCHAR(255) NOT NULL,
+            UM_EMAIL VARCHAR(255),
+            UM_ACTIVE BOOLEAN DEFAULT FALSE,
```
Nirothipan, line 5: Shall we format the script

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/SecurityHandlerAdapter.java
```diff
@@ -32,6 +44,27 @@
...
+        if (!isInitialized) {
+            ManagementApiParser mgtParser = new ManagementApiParser();
+            try {
+                usersList = mgtParser.getUserList();
+            } catch (UserStoreUndefinedException e) {
+                useCarbonUserStore = true;
+                LOG.info("User store config has not been defined in file " + ManagementApiParser.getConfigurationFilePath() + ". "
+                         + "Using carbon user store settings");
+            }
+            isInitialized = true;
```
Nirothipan, line 64: Shall we move this code block to a static method. Hence we can avoid updating static fields in the constructor.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/ManagementApiParser.java
```diff
@@ -0,0 +1,148 @@
...
+            }
+        }
+        throw new ManagementApiUndefinedException("Management API not defined in " + getConfigurationFilePath());
+    }
+
+    public HashMap<String, char[]> getUserList() throws CarbonException, XMLStreamException, IOException,
+            UserStoreUndefinedException, ManagementApiUndefinedException {
+        OMElement userStoreOM = getManagementApiElement().getFirstChildWithName(new QName("UserStore"));
+        if (Objects.nonNull(userStoreOM)) {
+            return populateUserList(userStoreOM.getFirstChildWithName(new QName("users")));
```
Nirothipan, line 89: Shall we hash this value and return, though we are initializing only once form SecurityHandlerAdapter in case if some other method uses this in future it will reduce the reads.

distribution/src/resources/config-tool/templates/conf/internal-apis.xml.j2
```diff
@@ -24,6 +24,18 @@
         <api name="{{api.api_name}}" protocol="{{api.protocols}}" class="{{api.class}}"/>
         {% endfor %}
         <api name="{{management_api.name}}" protocol="{{management_api.protocols}}" class="{{management_api.class}}">
+            {% if management_api.file_user_store.enable == true %}
```
Nirothipan, line 27: We need to update the documentation on this change

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UserResource.java
```diff
@@ -0,0 +1,158 @@
+/*
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
Nirothipan, line 2: Shall we update to 2020

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
```diff
@@ -0,0 +1,156 @@
+/*
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
Nirothipan, line 2: Shall we update the year

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
```diff
@@ -0,0 +1,111 @@
+/*
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
Nirothipan, line 2: Shall we update the year

distribution/src/resources/config-tool/templates/conf/internal-apis.xml.j2
Nirothipan: Also shall we look into the possibility of moving this user store outside the management api , i.e. common to all the apis so that we can include in other handler and use for other apis as well ?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
```diff
@@ -0,0 +1,111 @@
...
+import org.wso2.micro.integrator.security.MicroIntegratorSecurityUtils;
+import org.wso2.micro.integrator.security.user.api.UserStoreException;
+
+import javax.xml.stream.XMLStreamException;
+import java.io.IOException;
+import java.util.Arrays;
+
+import static org.wso2.micro.integrator.management.apis.Constants.USERNAME_PROPERTY;
+
+public class AuthorizationHandler extends SecurityHandlerAdapter {
```
Nirothipan, line 35: This handler will authenticate only /users call. Hence shouldn't we rename the class to reflect it .

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
```diff
@@ -0,0 +1,111 @@
...
+    @Override
+    public void setName(String name) {
+
+        this.name = name;
+    }
+
+    @Override
+    protected Boolean authenticate(String authHeaderToken) {
+
+        if ((messageContext.getTo().getAddress()).startsWith((Constants.REST_API_CONTEXT + Constants.PREFIX_USERS))) {
```
Nirothipan, line 67: Can we get "(Constants.REST_API_CONTEXT + Constants.PREFIX_USERS)" as a parameter for the handler from internal api xml, then we may use for other resources as well

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/SecurityHandlerAdapter.java
sdkottegoda: If we move it to a static method, we have no way of throwing an error while parsing. Also, we will not have a control over ordering if we move it to a static block. Therefore, I prefer leaving as it is. WDYT?

distribution/src/resources/config-tool/templates/conf/internal-apis.xml.j2
sdkottegoda: From the way it is now, we can use the user store from all the handlers added to the management API. I prefer not adding it above this since the rest of the internal APIs are read only APIs which do not require the user store. Therefore, I think it's better to wait until we get a requiremnt?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
sdkottegoda: I left it this way since we can use this for any resource which need authorization with a little modification.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
sdkottegoda: I thought about it, but didn't do since at the moment we do not have any other resources that require this capability. When, we get another let's do this.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/ManagementApiParser.java
sdkottegoda: done

distribution/src/resources/config-tool/default.json
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/ManagementApiParser.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UserResource.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
Nirothipan: Changing is the state of message processor is kind of write operation and people might tend to protect it .

distribution/src/resources/config-tool/templates/conf/internal-apis.xml.j2
Nirothipan: Recently we got a requirement to make protect Prometheus resource. I was having an idea in mind that if we move this out and make the resource path configurable for authentication, then we could use this handler for that also.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UserResource.java
```diff
@@ -0,0 +1,162 @@
...
+                case Constants.HTTP_GET: {
+                    response = handleGet(messageContext);
+                    break;
+                }
+                case Constants.HTTP_DELETE: {
+                    response = handleDelete(messageContext);
+                    break;
+                }
+                default: {
+                    response = Utils.createJsonError("Unsupported HTTP method, " + httpMethod + ". Only GET, POST, "
```
Nirothipan, line 86: The request won't reach here if the methods are not supported. It would be stopped in mediation. Hence we may leave this case empty or just use an else if.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UserResource.java
```diff
@@ -0,0 +1,162 @@
...
+        String[] roles = getUserStore().getRoleListOfUser(user);
+        userObject.put(IS_ADMIN, isAdmin(roles));
+        JSONArray list = new JSONArray(roles);
+        userObject.put(ROLES, list);
+        return userObject;
+    }
+
+    JSONObject handleDelete(MessageContext messageContext) throws UserStoreException, IOException,
+            ResourceNotFoundException {
+        LOG.info("Handling DELETE");
```
Nirothipan, line 119: This needs to be a debug log ?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UserResource.java
```diff
@@ -0,0 +1,162 @@
...
+
+    @Override
+    public Set<String> getMethods() {
+        return methods;
+    }
+
+    @Override
+    public boolean invoke(MessageContext messageContext, org.apache.axis2.context.MessageContext axis2MessageContext,
+                          SynapseConfiguration synapseConfiguration) {
+        LOG.info("Handling request by user resource");
```
Nirothipan, line 72: Should we print info log for each invocation ? and if so we need to print with more information like the user who is invoking this may be

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
```diff
@@ -0,0 +1,110 @@
...
+            if (useCarbonUserStore) {
+                //Uses carbon user store
+                try {
+                    return processAuthorizationWithCarbonUserStore();
+                } catch (UserStoreException e) {
+                    LOG.error("Error while authenticating with carbon user store", e);
```
Nirothipan, line 73: we need to return false here else it will come out of if block and will return true

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
```diff
@@ -0,0 +1,110 @@
...
+                }
+            } else {
+                //Uses in memory user store
+                LOG.error("User management is not supported with the in memory user store. Users need to be modified "
```
Nirothipan, line 77: This handler is about authorization and not management , isn't it .. In that case, is this log correct

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UserResource.java
sdkottegoda: If by any chance the request is forwarded upto this point, it'll throw this error, I see no harm in having it. Else if will complicate things if a new method gets introduced.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UserResource.java
sdkottegoda: no, mistake :D

distribution/src/resources/config-tool/templates/conf/internal-apis.xml.j2
sdkottegoda: Let's do this as a separate issue, this PR is big enough already. https://github.com/wso2/micro-integrator/issues/1450 :)

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UserResource.java
Nirothipan: Yes no harm in having this . I was thinking that since this can't reach here logically and if we remove this, code will look simple and we can handle that erroneous scenario in a test case may be . Both are fine.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/ManagementApiParser.java
```diff
@@ -0,0 +1,154 @@
...
+        HashMap<String, char []> userList = new HashMap<>();
+        if (users != null) {
+            Iterator usersIterator = users.getChildrenWithName(new QName("user"));
+            if (usersIterator != null) {
+                while (usersIterator.hasNext()) {
+                    OMElement userElement = (OMElement) usersIterator.next();
+                    OMElement userNameElement = userElement.getFirstChildWithName(new QName("username"));
+                    OMElement passwordElement = userElement.getFirstChildWithName(new QName("password"));
+                    if (userNameElement != null && passwordElement != null) {
+                        userList.put(userNameElement.getText(), passwordElement.getText().toCharArray());
```
arunans23, line 137: Do we need to validate whether same userName is defined twice?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UserResource.java
arunans23, line 86: Is POST supported? Log says so

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
```diff
@@ -0,0 +1,150 @@
...
+import static org.wso2.micro.integrator.management.apis.Constants.IS_ADMIN;
+import static org.wso2.micro.integrator.management.apis.Constants.LIST;
+import static org.wso2.micro.integrator.management.apis.Constants.PASSWORD;
+import static org.wso2.micro.integrator.management.apis.Constants.PATTERN;
+import static org.wso2.micro.integrator.management.apis.Constants.ROLE;
+import static org.wso2.micro.integrator.management.apis.Constants.STATUS;
+import static org.wso2.micro.integrator.management.apis.Constants.USER_ID;
+
+/**
+ * Resource for login. This obtains JWT token. This is basic auth protected.
```
arunans23, line 47: Shall we differentiate between UserResource and UsersResource in this javadoc?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
```diff
@@ -0,0 +1,150 @@
...
+                case Constants.HTTP_GET: {
+                    response = handleGet(messageContext);
+                    break;
+                }
+                case Constants.HTTP_POST: {
+                    response = handlePost(axis2MessageContext);
+                    break;
+                }
+                default: {
+                    response = Utils.createJsonError("Unsupported HTTP method, " + httpMethod + ". Only GET, POST, "
```
arunans23, line 77: Seems DELETE is not supported. Shall we fix the message?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
```diff
@@ -0,0 +1,150 @@
...
+            String user = payload.get(USER_ID).getAsString();
+            getUserStore().addUser(user, payload.get(PASSWORD).getAsString(),
+                                   roleList, null, null);
+
+            JSONObject jsonBody = new JSONObject();
+            jsonBody.put(USER_ID, user);
+            jsonBody.put(STATUS, "Added");
+            return jsonBody;
+        } else {
+            throw new IOException("Missing one or more of the fields, '" + USER_ID + "', '" + PASSWORD + "' in the "
```
arunans23, line 145: Why are we throwing an IOException here and catching it in the invoke method? Can we look into a better way?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/JWTConfig.java
```diff
@@ -64,116 +59,61 @@ public static JWTConfig getInstance() {
...
+                        if (Objects.nonNull(removeOldestElem)) {
+                            jwtDTO.setRemoveOldestElementOnOverflow(Boolean.parseBoolean(removeOldestElem.getText()));
+                        }
+                        OMElement cleanupIntervalElem = tokenStoreConfigOM.getFirstChildWithName(new QName(
+                                "TokenCleanupTaskInterval"));
+                        if (Objects.nonNull(cleanupIntervalElem)) {
+                            jwtDTO.setCleanupThreadInterval(Integer.parseInt(cleanupIntervalElem.getText()));
+                        }
+                    } else {
+                        LOG.fatal("Token Store config has not been defined in file " + ManagementApiParser.getConfigurationFilePath() + " Using default values");
```
arunans23, line 87: Let's break the line here

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/ManagementApiParser.java
Nirothipan: I couldn't find an assignment for the hashed value.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/SecurityHandlerAdapter.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/ManagementApiParser.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UserResource.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UserResource.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/ManagementApiParser.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/JWTConfig.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
sdkottegoda: What's the issue with doing it this way? For this method, it is an error in the request payload hence an IO Exception, and for the invoke method, an IO Exception is a bad request. I thought of it that way. Do you see any issue with that?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UserResource.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
arunans23: I think it is fine. I only considered this part of the method. But with a higher picture, it seems reasonable.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
sdkottegoda: done

sdkottegoda:
This PR does not contain the templates for the internal-apis.xml and and the implementation for making user store common to all apis. Those will be handles separately once this is merged.

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
```diff
@@ -199,17 +212,19 @@ private static void populateHandlers(OMElement apiElement, InternalAPI api) {
...
 
-    private static InternalAPIHandler createHandler(String classFQName) {
+    private static InternalAPIHandler createHandler(String classFQName, String context, List<String> resources) {
 
         try {
-            Object obj = Class.forName(classFQName).newInstance();
+            Constructor c = Class.forName(classFQName).getConstructor(String.class);
+            Object obj = c.newInstance(context);
             if (obj instanceof InternalAPIHandler) {
+                ((InternalAPIHandler) obj).setResources(resources);
```
Nirothipan, line 221: Shall we assign the casted object to a variable and use it as we are again casting while returning.

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
```diff
@@ -199,17 +212,19 @@ private static void populateHandlers(OMElement apiElement, InternalAPI api) {
...
                 return (InternalAPIHandler) obj;
             } else {
                 throw new SynapseException("Error creating Internal InternalAPIHandler. "
                                                    + "The InternalAPIHandler should be of type InternalAPIHandler");
             }
-        } catch (ClassNotFoundException | InstantiationException | IllegalAccessException e) {
+        } catch (ClassNotFoundException | InstantiationException | IllegalAccessException | NoSuchMethodException | InvocationTargetException e) {
```
Nirothipan, line 227: Shall we format this as this exceeds 120 charecters.

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/InternalAPIHandler.java
```diff
@@ -45,4 +47,15 @@
...
      */
     Boolean invoke(MessageContext synCtx);
+
+    /**
+     * Sets the resources for the handler to be applied to.
+     */
+    void setResources(List<String> resources);
+
+    /**
+     * Sets the resources for the handler to be applied to.
```
Nirothipan, line 57: This should be get : )

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
```diff
@@ -0,0 +1,170 @@
...
+                    response = handleGet(messageContext);
+                    break;
+                }
+                case Constants.HTTP_POST: {
+                    response = handlePost(axis2MessageContext);
+                    break;
+                }
+                default: {
+                    response = Utils.createJsonError("Unsupported HTTP method, " + httpMethod + ". Only GET and "
+                                                     + "DELETE,  methods are supported",
```
Nirothipan, line 82: Only GET and POST , isn't it ?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
```diff
@@ -0,0 +1,170 @@
...
+
+    protected JSONObject handleGet(MessageContext messageContext) throws UserStoreException {
+        String searchPattern = Utils.getQueryParameter(messageContext, PATTERN);
+        if (Objects.isNull(searchPattern)) {
+            searchPattern = "*";
+        }
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("Searching for users with the pattern: " + searchPattern);
+        }
+        List<String> patternUsersList = Arrays.asList(getUserStore().listUsers(searchPattern, -1));
```
Nirothipan, line 106: Shall we move this line on top of the above debug log so that we can combine both debugs and don't need to check two times that debug is enabled.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
```diff
@@ -0,0 +1,170 @@
...
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("Retrieved list of users for the pattern: ");
+            patternUsersList.forEach(LOG::debug);
+        }
+        String roleFilter = Utils.getQueryParameter(messageContext, ROLE);
```
Nirothipan, line 111: Same here.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
```diff
@@ -0,0 +1,170 @@
...
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("Filtered list of users: ");
+            users.forEach(LOG::debug);
+        }
+        jsonBody = Utils.createJSONList(users.size());
+        for (String user : users) {
+            JSONObject userObject = new JSONObject();
+            userObject.put(USER_ID, user);
+            jsonBody.getJSONArray(LIST).put(userObject);
+
```
Nirothipan, line 136: Shall we remove this empty line

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/Utils.java
```diff
@@ -55,6 +55,20 @@ public static String getQueryParameter(MessageContext messageContext, String key
...
+    /**
+     * Extracts the value set for the patch parameter with the given key.
+     *
+     * @param messageContext message context to extract the parameter from
+     * @param key            the key defined in the uri template
+     * @return the resolved value from the url. Returns null if not present.
+     */
+    public static String getPathParameter(MessageContext messageContext, String key){
+        if (Objects.nonNull(messageContext.getProperty(RESTConstants.REST_URI_VARIABLE_PREFIX + key))) {
+            return messageContext.getProperty(RESTConstants.REST_URI_VARIABLE_PREFIX + key).toString();
```
Nirothipan, line 67: Shall we assign "messageContext.getProperty(RESTConstants.REST_URI_VARIABLE_PREFIX + key)" to a variable and do the processing without retrieving from the msg context two times.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
sdkottegoda: I added the 2 logs before and after searching the users from the store, since something could go wrong there.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
sdkottegoda: same as above.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/Utils.java
```diff
@@ -76,6 +90,45 @@ public static JSONObject createJSONList(int count) {
         return jsonBody;
     }
 
+    /**
+     * Creates a json response according to the message provided and sets the provided HTTP code.
+     *
+     * @param message             the error response to be sent to the client
+     * @param exception           the exception to be logged on the server side. The error resone will be extracted
```
Nirothipan, line 97:
```suggestion
     * @param exception           the exception to be logged on the server side. The error response will be extracted
```

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
```diff
@@ -0,0 +1,119 @@
...
+import org.wso2.micro.integrator.security.user.api.UserStoreException;
+
+import javax.xml.stream.XMLStreamException;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import static org.wso2.micro.integrator.management.apis.Constants.USERNAME_PROPERTY;
+
+public class AuthorizationHandler extends SecurityHandlerAdapter {
```
Nirothipan, line 36: Shall we add a small comment describing how this works

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
sdkottegoda: done

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
sdkottegoda: done

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/InternalAPIHandler.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/UsersResource.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/Utils.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/Utils.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
```diff
@@ -0,0 +1,119 @@
...
+        } else {
+            //Uses in memory user store
+            LOG.warn("Authorization is not supported with the in memory user store. The request will be authorized. "
+                     + "Please plug in a user store for the correct functionality");
+            return true;
+        }
+    }
+
+    /**
+     * Processes /users request if the user is an admin
```
Nirothipan, line 91:
```suggestion
     * Processes /users request if the user is an admin .
```

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
```diff
@@ -0,0 +1,119 @@
...
+                return processAuthorizationWithCarbonUserStore();
+            } catch (UserStoreException e) {
+                LOG.error("Error while authenticating with carbon user store", e);
+                return false;
+            }
+        } else {
+            //Uses in memory user store
+            LOG.warn("Authorization is not supported with the in memory user store. The request will be authorized. "
+                     + "Please plug in a user store for the correct functionality");
+            return true;
```
Nirothipan, line 86: Shouldn't we log an error and return false. Warn might get un noticed and the one who used it incorrectly might thinking that it is authenticated.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/BasicSecurityHandler.java
```diff
@@ -49,7 +41,12 @@
 
     private String name;
     private SecretResolver secretResolver;
```
Nirothipan, line 43: Shall we remove this variable, as we have removed the usage of it.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/BasicSecurityHandler.java
```diff
@@ -63,7 +60,9 @@ public void setName(String name) {
 
     @Override
     protected Boolean authenticate(String authHeaderToken) {
-
+        if (LOG.isDebugEnabled()) {
```
Nirothipan, line 63: Shall we remove this if check just for the code to be simple.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/JWTConfig.java
```diff
@@ -64,116 +59,62 @@ public static JWTConfig getInstance() {
...
+                        OMElement sizeElem = tokenConfigOM.getFirstChildWithName(new QName("size"));
+                        OMElement expiryElem = tokenConfigOM.getFirstChildWithName(new QName("expiry"));
+                        if (Objects.nonNull(expiryElem)) {
+                            jwtDTO.setExpiry(expiryElem.getText());
+                        }
+                        if (Objects.nonNull(sizeElem)) {
+                            jwtDTO.setTokenSize(sizeElem.getText());
+                        }
+                    } else {
+                        LOG.fatal("Token config has not been defined in file " + ManagementApiParser.getConfigurationFilePath() + " Using default values");
```
Nirothipan, line 101: Shall we break the line here.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/JWTTokenSecurityHandler.java
```diff
@@ -21,19 +21,28 @@
...
 
     private static final Log LOG = LogFactory.getLog(JWTTokenSecurityHandler.class);
     private String name;
-    private Map<String, char[]> userList = null;
-    private MessageContext messageContext;
+
+    public JWTTokenSecurityHandler(String context) throws CarbonException, XMLStreamException, IOException,
+            ManagementApiUndefinedException {
+        super(context);
+        populateDefaultResources();
```
Nirothipan, line 44: Shall we move this to super class as this is common to all IINM

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/test/java/org/wso2/micro/integrator/management/apis/ManagementApiParserTest.java
```diff
@@ -0,0 +1,77 @@
...
+        Assert.assertNotNull(userList.get("user1"));
+        Assert.assertEquals("pwd1", String.valueOf(userList.get("user1")));
+
+        //Assert user2:pwd2
+        Assert.assertNotNull(userList.get("user2"));
+        Assert.assertEquals("pwd2", String.valueOf(userList.get("user2")));
+    }
+
+
+}
```
Nirothipan, line 77: Shall we add a line at EOF

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/test/resources/org/wso2/micro/integrator/management/apis/internal-apis.xml
```diff
@@ -0,0 +1,42 @@
+<!--
+  ~ Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
Nirothipan, line 2: Shall we update the year

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/BasicSecurityHandler.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/BasicSecurityHandler.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/JWTConfig.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/JWTTokenSecurityHandler.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/test/resources/org/wso2/micro/integrator/management/apis/internal-apis.xml
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/test/java/org/wso2/micro/integrator/management/apis/ManagementApiParserTest.java
sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
```diff
@@ -0,0 +1,123 @@
...
+            try {
+                return processAuthorizationWithCarbonUserStore();
+            } catch (UserStoreException e) {
+                LOG.error("Error while authenticating with carbon user store", e);
+                return false;
+            }
+        } else {
+            //Uses in memory user store
+            LOG.error("Authorization is not supported with the in memory user store. The request will be authorized. "
+                      + "Please plug in a user store for the correct functionality");
```
Nirothipan, line 89: This should be , please plugin in a JDBC user store ?

sdkottegoda: done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandler.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /**
+     * Processes /users request if the user is an admin.
+     *
+     * @return if successfully authorized
+     */
+    private boolean processAuthorizationWithCarbonUserStore() throws UserStoreException {
+
+        String username = messageContext.getProperty(USERNAME_PROPERTY).toString();
```
Nirothipan, line 101: Since we have the access token in header, can't we get the user name from the JWT Token Store , else this handler will always be coupled with JWT handler and won't be able to use this without JWT even if we have the token.

sdkottegoda: If we take it from the JWT token store, this handler will be coupled with JWT. The way it is written now, it can be used with any authentication handler. It is mandatory to have some kind of an authentication handler prior to using the authorization handler since, otherwise it will be doing the authentication part aswell.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/security/handler/AuthorizationHandlerAdapter.java
```diff
@@ -0,0 +1,64 @@
...
+    private static final Log LOG = LogFactory.getLog(AuthorizationHandlerAdapter.class);
+
+    public AuthorizationHandlerAdapter(String context) throws CarbonException, XMLStreamException, IOException,
+            ManagementApiUndefinedException {
+        super(context);
+    }
+
+    @Override
+    public Boolean handle(MessageContext messageContext) {
+        String userName = messageContext.getProperty(USERNAME_PROPERTY).toString();
```
Nirothipan, line 47: This would lead to NPE if its nto there. we can check only null and return false.

sdkottegoda: done

PR #1456, opened 2020-04-23
Template synapse-handler configurations in deployment.toml file

sajinieKavindya:
## Purpose
Template synapse-handler configurations in deployment.toml file as in the following example.

Ex:
[[synapse_handlers]]
name="IntegratorHandler"
class="org.wso2.micro.integrator.example.SynapseHandler"


Nirothipan:
Let's add this to docs as well.

PR #1493, opened 2020-04-28
Add exists and create operations to mongo dataservice queries

Drifftr:
## Purpose
Add exists and create operations to mongo dataservice queries


chanikag:
Shall we do the same fix in carbon-data as well?

Drifftr:
Sure will do

Drifftr:
Added same fix for carbon-data - https://github.com/wso2/carbon-data/pull/296

PR #1494, opened 2020-04-28
Disable Mail transport by default

sdkottegoda:


Nirothipan:
Need to check whether we need a doc udpate .

sdkottegoda:
> Need to check whether we need a doc udpate .

The documentation has the correct configurations

PR #1502, opened 2020-04-29
Fix issue of unit tests not being run

sdkottegoda:
## Purpose
The unit tests were not running because they were inside src -> main -> test while they need to be inside src -> test. Moving the test sources to the correct directory fixes the issue.


components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/test/java/org/wso2/micro/integrator/management/apis/ManagementApiParserTest.java
```diff
@@ -33,13 +33,6 @@
 
 public class ManagementApiParserTest {
 
-    @Test
```
sdkottegoda, line 36: This is removed because in Windows we are expecting (getClass().getResource("internal-apis.xml").getPath()) the absolute path to start with a "/". e.g., "/D:/somepath". Even though it is possible to evaluate it without the "/" it doesn't apply to all the operating systems, hence it sounds like useless. In addition, this is just an assertion to check if the resource file was correctly located, not exactly a test to test the functionality of the Parser.

PR #1518, opened 2020-05-01
Fix Intermittent test failure

Nirothipan:
$subject. The tests were failing since same seq is used in more than single test case and when the order changed, it lead to failures. Hence separated them out. Also the \n character and spaces in payload was causing un expected results when checking string contains and etc.

In case of Message Processor tests, we were asserting a generic string which is not common to this test case, Hence modified the assert message specific to this test case so that it don't conflict with others. Also in an assert, three conditions were tested. So break them into three so that we can identify the failures more clearly if it fails in future.


sdkottegoda:
Shall we mention the cause of the test failure and the solution please.

Nirothipan:
> Shall we mention the cause of the test failure and the solution please.

Updated the description.

PR #1528, opened 2020-05-05
Fix Test Failure

Nirothipan:
$subject.

This test case looks for a log "**in_payload**" which is not printed by the proxy which being invoked in the test case. If by any chance, another test case executes in between, and prints the particular then this test case validates all the conditions inside that if block. Hence removed that if check. 

When it validate, since the modified payload by the for each mediator is not logged, all the if conditions fail. Hence added a log level full.

### DIFF
--- a/integration/mediation-tests/tests-mediator-1/src/test/java/org/wso2/carbon/esb/mediator/test/foreach/ForEachPropertiesTestCase.java
+++ b/integration/mediation-tests/tests-mediator-1/src/test/java/org/wso2/carbon/esb/mediator/test/foreach/ForEachPropertiesTestCase.java
@@ -101,28 +101,21 @@ public class ForEachPropertiesTestCase extends ESBIntegrationTest {
             assertTrue(carbonLogReader.getLogs().contains("in_count = " + 3),
                     "Final counter mismatch, expected 3 found = " + carbonLogReader.getLogs());
         }
+        //final payload in insequence
+        String payload = carbonLogReader.getLogs();
+        String search = "<m0:getQuote>(.*)</m0:getQuote>";
+        Pattern pattern = Pattern.compile(search, Pattern.DOTALL);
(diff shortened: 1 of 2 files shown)


integration/mediation-tests/tests-mediator-1/src/test/java/org/wso2/carbon/esb/mediator/test/foreach/ForEachPropertiesTestCase.java
```diff
@@ -101,28 +101,21 @@ public void testSingleForEachProperties() throws Exception {
...
+        boolean matchFound = matcher.find();
+
+        assertTrue(matchFound, "getQuote element not found");
+        int start = matcher.start();
+        int end = matcher.end();
+        String quote = payload.substring(start, end);
+        assertTrue(quote.contains("<m0:group>Group1</m0:group>"), "Group Element not found in : " + quote);
+        assertTrue(quote.contains("<m0:symbol>Group1_IBM</m0:symbol>"), "IBM Element not found in : " + quote);
+        assertTrue(quote.contains("<m0:symbol>Group1_WSO2</m0:symbol>"), "WSO2 Element not found in : " + quote);
+        assertTrue(quote.contains("<m0:symbol>Group1_MSFT</m0:symbol>"), "MSTF Element not found in : " + quote);
```
sdkottegoda, line 118: I see that we have only removed the outer if condition. Even though we've removed the inner if (matchFound) {, we still have the assert, which I feel like is still the same as before. Am I missing something?

Nirothipan: Yes the inner asserts will fail if it reach there since we are not logging the payload. Hence added a log level full in the artifact to fix this case as mentioned in the PR description.

Issue #1490, opened 2020-04-27
Add mechanism to count number of requests handled by the server


sdkottegoda:
**Description:**

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Related Issues:**

sachithKay:
Closing the issue with PR: https://github.com/wso2/micro-integrator/pull/1537

PR #1537, opened 2020-05-11
Add Transaction counting mechanism

sajinieKavindya:
## Purpose
This PR adds the functionality to count the number of transactions come to the WSO2-MI server and persist the count in the database configured in deployment.toml. 

Related issue: https://github.com/wso2/micro-integrator/issues/1490


components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
```diff
@@ -0,0 +1,172 @@
...
+     * Get the time rate for running the transaction writer scheduled task.
+     *
+     * @return - scheduled period.
+     */
+    private int getTransactionCountScheduledPeriod() {
+        String scheduledPeriodStr = System.getProperty(SCHEDULED_PERIOD);
+        if (StringUtils.isEmpty(scheduledPeriodStr)) {
+            scheduledPeriodStr = System.getenv(SCHEDULED_PERIOD);
+        }
+        if (StringUtils.isEmpty(scheduledPeriodStr)) {
```
Nirothipan, line 82: Shall we move this inside above if block.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Get the dataSource configured for transaction counting utility.
+     *
+     * @return - datasource id.
+     * @throws TransactionException - when a dataSource is not configured for transaction counting utility.
+     */
+    private String getTransactionDatasourceId() throws TransactionException {
+        String dataSourceId = System.getProperty(TRANSACTION_DATA_SOURCE);
+        if (StringUtils.isEmpty(dataSourceId)) {
+            dataSourceId = System.getenv(TRANSACTION_DATA_SOURCE);
```
Nirothipan, line 149: Do we need this much of flexibility of reading this from system env .. Doesn't toml only suffice.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
```diff
@@ -0,0 +1,172 @@
...
+            if (StringUtils.isEmpty(dataSourceId)) {
+                Object dataSourceIdObject = ConfigParser.getParsedConfigs().get(
+                        TRANSACTION_CONFIG_SECTION + "." + TRANSACTION_CONFIG_NAME);
+                if (dataSourceIdObject != null) {
+                    dataSourceId = dataSourceIdObject.toString();
+                } else {
```
Nirothipan, line 155: Shall we default to carbon db if not configured.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
```diff
@@ -0,0 +1,172 @@
...
+            DataSource dataSource = getTransactionDataSource(dataSourceService);
+
+            // initialize Transaction store.
+            TransactionStore transactionStore = new TransactionStore(dataSource, nodeId, cipher);
+
+            // start and schedule transaction writer task.
+            scheduleTransactionWriterTask(transactionStore);
+
+        } catch (DataSourceException | TransactionException e) {
+            LOG.error("Error in initializing Transaction Service component: " + e.getMessage(), e);
```
Nirothipan, line 68: Shall we throw this and stop the server

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/store/TransactionStore.java
```diff
@@ -0,0 +1,55 @@
...
+import org.wso2.micro.integrator.initializer.handler.transaction.TransactionException;
+import org.wso2.micro.integrator.initializer.handler.transaction.store.connector.RDBMSConnector;
+
+import javax.crypto.Cipher;
+import javax.sql.DataSource;
+
+/**
+ * The layer which connects to the transaction data.
+ */
+public class TransactionStore {
```
Nirothipan, line 30: Shall we make this as an interface and have an RDBMS store , we can use file based store for stand alone deployments. 
ELse how are we going to handle one only one node is there , is db a must ?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sajinieKavindya: Done.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sajinieKavindya: Even though this feature is enabled by default, we are not strictly enforcing the users to have this feature enabled. So IMO, warning or an error log would be enough to notify the users who do not want this feature and forget to disable it (or worst case, want the feature but forget to configure a dataSource). WDYT?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/store/TransactionStore.java
sajinieKavindya: In the Kubernetes environment, each pod will consider itself as a standalone deployment. In that case, when we have a file-based store for standalone deployments and unless this feature is disabled, each pod/container will continue to store the count inside the pod where the data is not persistent. This can be an unnecessary overhead where the stored data has no actual use. How can we address that?

Nirothipan: True for Kubernetes. I was having VM standlone deployments in mind, how are we going to address it there

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
Nirothipan: I think if it is enabled by default , we will be having some storage by default and under normal circumstances this would work flawlessly. If the user edit some configs and having in mind that the transactions will be counted, IMO opinion it would be not that correct to print an error and proceed and later they figure out that their transaction count is missing which is irrecoverable.

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSInjectHandler.java
```diff
@@ -120,6 +121,8 @@ public boolean invoke(Object object, String name) throws SynapseException {
             //setting transport headers
             axis2MsgCtx.setProperty(org.apache.axis2.context.MessageContext.TRANSPORT_HEADERS,
                                     JMSUtils.getTransportHeaders(msg, axis2MsgCtx));
+            // set transaction property
+            axis2MsgCtx.setProperty(BaseConstants.TRANSACTION, msg.getStringProperty(BaseConstants.TRANSACTION));
```
sdkottegoda, line 125: Does the property need to be a String? Can it be a boolean?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
```diff
@@ -165,6 +175,20 @@ protected void activate(ComponentContext ctxt) {
...
                 synapseEnvironment.registerSynapseHandler(new SynapseExternalPropertyConfigurator());
                 synapseEnvironment.registerSynapseHandler(new ProxyLogHandler());
+
+                // Register internal transaction synapse handler
+                boolean transactionPropertyEnabled = TransactionHandlerComponent.isTransactionPropertyEnabled();
+                if (transactionPropertyEnabled) {
+                    TransactionHandler transactionHandler = new TransactionHandler();
+                    synapseEnvironment.registerSynapseHandler(transactionHandler);
+                    try {
+                        transactionHandlerComponent = new TransactionHandlerComponent();
```
sdkottegoda, line 185: Will the word requestCountHandler/TransactionCountHandler be clear?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
```diff
@@ -165,6 +175,20 @@ protected void activate(ComponentContext ctxt) {
 
                 synapseEnvironment.registerSynapseHandler(new SynapseExternalPropertyConfigurator());
                 synapseEnvironment.registerSynapseHandler(new ProxyLogHandler());
+
+                // Register internal transaction synapse handler
+                boolean transactionPropertyEnabled = TransactionHandlerComponent.isTransactionPropertyEnabled();
+                if (transactionPropertyEnabled) {
+                    TransactionHandler transactionHandler = new TransactionHandler();
```
sdkottegoda, line 182: Will the word requestCountHandler/TransactionCountHandler be clear?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
```diff
@@ -165,6 +175,20 @@ protected void activate(ComponentContext ctxt) {
...
+                    synapseEnvironment.registerSynapseHandler(transactionHandler);
+                    try {
+                        transactionHandlerComponent = new TransactionHandlerComponent();
+                        transactionHandlerComponent.start(dataSourceService);
+                    } catch (DataSourceException | TransactionException e) {
+                        log.warn("Error in initializing Transaction Service component: " + e.getMessage(), e);
```
sdkottegoda, line 188: Would the level ERROR be better since we are logging an error anyway. Also , we don't need to log e.getMessage() it'll be logged when you have the exception as a parameter

sdkottegoda: Also, now that anybody who does not have a datasource configured will see this message, I think it should be a bit more descriptive. "Error in initializing the Transaction component. The number of transactions will not be recorded. You can ignore this error if you don't want the server to keep track of the number of transactions processed"

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
```diff
@@ -194,7 +218,9 @@ protected void activate(ComponentContext ctxt) {
 
     @Deactivate
     protected void deactivate(ComponentContext ctxt) {
-
+        if (Objects.nonNull(transactionHandlerComponent)) {
```
sdkottegoda, line 221: Can this ever be null?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
```diff
@@ -96,6 +102,10 @@
 
     private TaskService taskService;
 
+    private static DataSourceService dataSourceService;
```
sdkottegoda, line 105: Should this be a static?

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSUtils.java
```diff
@@ -317,6 +318,7 @@ protected static void handleException(String s) throws NamingException {
                     map.put(headerName, headerValue);
                 }
             }
+            map.remove(BaseConstants.TRANSACTION);
```
sdkottegoda, line 321: Why are we removing this in the case of an exception? Shall we have a small comment explaining it.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionConstants.java
```diff
@@ -0,0 +1,33 @@
...
+
+package org.wso2.micro.integrator.initializer.handler.transaction;
+
+public class TransactionConstants {
+
+    public static final String CARBON_HOME = "carbon.home";
+    public static final String REPOSITORY_DIR = "repository";
+    public static final String RESOURCES_DIR = "resources";
+    public static final String SECURITY_DIR = "security";
+    public static final String KEYSTORE_FILE = "wso2carbon.jks";
```
sdkottegoda, line 27: I think we already have constants for these

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandler.java
```diff
@@ -0,0 +1,60 @@
...
+public class TransactionHandler extends AbstractSynapseHandler {
+
+    private static long transactionCount = 0;
+
+    @Override
+    public boolean handleRequestInFlow(MessageContext messageContext) {
+        org.apache.axis2.context.MessageContext axis2MessageContext =
+                ((Axis2MessageContext) messageContext).getAxis2MessageContext();
+        Object transactionProperty = axis2MessageContext.getProperty(BaseConstants.TRANSACTION);
+        if (!BaseConstants.TRANSACTION_COUNTED.equals(transactionProperty)) {
```
sdkottegoda, line 35: Is it possible to have INTERNAL_TRANSACTION_COUNT as the property name and true/false as its values. This will be a boolean property? Do you think it is ok?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
abeykoon: +1. `TransactionHandler` will mean a component actually handling the transactions.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
```diff
@@ -0,0 +1,189 @@
...
+
+    private ScheduledExecutorService trnCountWriterTaskScheduler;
+
+    public void start(DataSourceService dataSourceService) throws TransactionException, DataSourceException {
+        String nodeId = generateRandomId();
+
+        // initialize cipher for encryption needs.
+        Cipher cipher = CryptoUtil.initializeCipher();
+
+        DataSource dataSource = getTransactionDataSource(dataSourceService);
```
sdkottegoda, line 61: I think we need to initialize the datasource before anything else, since if that is not present we should not be doing anything here. WDYT?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
```diff
@@ -96,6 +102,10 @@
...
+
+    private TransactionHandlerComponent transactionHandlerComponent;
```
abeykoon, line 107: What abt `TansactionCountHandler`? IMO, `TransactionHandler` itself gives a broader idea. We are not handling transactions. We just count here. Isn't it?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionException.java
```diff
@@ -0,0 +1,30 @@
...
+* software distributed under the License is distributed on an
+* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+* KIND, either express or implied. See the License for the
+* specific language governing permissions and limitations
+* under the License.
+*/
+
+package org.wso2.micro.integrator.initializer.handler.transaction;
+
+public class TransactionException extends Exception {
```
abeykoon, line 21: Shall we modify the exception name as well - `TransactionCounterException`?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
```diff
@@ -0,0 +1,189 @@
...
+        scheduleTransactionWriterTask(transactionStore);
+    }
+
+    /**
+     * Get the time rate for running the transaction writer scheduled task.
+     *
+     * @return - scheduled period.
+     */
+    private int getTransactionCountScheduledPeriod() {
+        String scheduledPeriodStr = System.getProperty(SCHEDULED_PERIOD);
```
abeykoon, line 76: We need to document this priority oder. Pls. note.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
```diff
@@ -0,0 +1,189 @@
...
+                Object scheduledPeriodObject = ConfigParser.getParsedConfigs().get("" + "." + "");
+                if (null != scheduledPeriodObject) {
+                    scheduledPeriodStr = scheduledPeriodObject.toString();
+                }
+            }
+        }
+        try {
+            return Integer.parseInt(scheduledPeriodStr);
+        } catch (NumberFormatException e) {
+            return DEFAULT_SCHEDULED_PERIOD_INTERVAL;
```
abeykoon, line 89: Should we issue a WARN here? User input is wrong. We should not silently failover to default value.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+    }
+
+    /**
+     * Schedule the transaction writer task to write the current transaction count to the database at a fixed rate.
+     *
+     * @param transactionStore - transactionStore instance.
+     */
+    private void scheduleTransactionWriterTask(TransactionStore transactionStore) {
+        trnCountWriterTaskScheduler = Executors.newSingleThreadScheduledExecutor();
```
abeykoon, line 99: txCountWriterTaskScheduler. It is usual to use `tx` for `transaction`. Just a suggestion.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
```diff
@@ -0,0 +1,189 @@
...
+        int scheduledPeriod = getTransactionCountScheduledPeriod();
+
+        trnCountWriterTaskScheduler.scheduleAtFixedRate(() -> {
+            try {
+                transactionStore.addTransaction();
```
abeykoon, line 104: Could there be any other type of exceptions? We need to make sure, this task will not get killed due to any reason (exception throw). It should log error and continue running.

abeykoon: Shall we add a debug log here?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/security/CryptoUtil.java
```diff
@@ -0,0 +1,122 @@
...
+        KeyStore primaryKeyStore = getKeyStore(getAbsolutePathToKeyStoreLocation(),
+                                               TransactionConstants.KEYSTORE_PASSWORD,
+                                               TransactionConstants.KEY_TYPE);
+        try {
+            Certificate certs = primaryKeyStore.getCertificate(TransactionConstants.KEY_ALIAS);
+            cipher = Cipher.getInstance(TransactionConstants.ENCRYPTION_ALGORITHM);
+            cipher.init(Cipher.ENCRYPT_MODE, certs);
+            LOG.debug("Cipher for Transaction Handler component is initialized Successfully");
+        } catch (KeyStoreException | NoSuchAlgorithmException | NoSuchPaddingException | InvalidKeyException e) {
+            throw new TransactionException("Error initializing Cipher ", e);
```
abeykoon, line 72: Cipher for what? EI can use ciphers for many different things.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/store/TransactionStore.java
abeykoon: VM standalone deployments won't have a DB. Shall we use H2 based carbon DB in that case? If a RDBMS is not present, if counting is enabled, it should default to carbon DB.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
sajinieKavindya: Changed.

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSUtils.java
sajinieKavindya: Done.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandler.java
sajinieKavindya: Done.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sajinieKavindya: Done.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionException.java
sajinieKavindya: +1. Changed.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
```diff
@@ -0,0 +1,189 @@
...
+import javax.sql.DataSource;
+
+/**
+ * This class represents the internal transaction count handler component.
+ */
+public class TransactionHandlerComponent {
+
+    private static final Log LOG = LogFactory.getLog(TransactionHandlerComponent.class);
+    private static final int DEFAULT_SCHEDULED_PERIOD_INTERVAL = 60;
+    private static final String SCHEDULED_PERIOD = "transactionCountScheduledPeriod";
```
Nirothipan, line 47: Shall we have it update period or something to reflect the real stuff it is doing. For an end user schedule period won't be meaningful IMO

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sajinieKavindya: +1. Added a warn log.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sdkottegoda: we cannot default to carbon db since we do not expect everybody to configure it. Not even a majority. We thought of the coordination db, but once again, we cannot expect everybody to have that either.

If the user doesn't have this configuration, it is highly likely that he hasn't created a database tables either. Since this is a conscious thing he has to do, we thought of no defaulting it, WDYT?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sdkottegoda: I think we do not need to read from the env/sys variables since, it is highly unlikely that they might change across environments

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sdkottegoda: I think we should log an error and continue, if a datasource is not found. If it is located, we need to stop the server for all the subsequent errors

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
sdkottegoda: +1

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sajinieKavindya: changed to "update_interval"

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sajinieKavindya: Then, shall we catch the "Exception" here? WDYT?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sajinieKavindya: +1

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sajinieKavindya: Changed the code to provide the configuration option of the update interval only via the deployment.toml.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
sajinieKavindya: +1

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
sajinieKavindya: +1

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/security/CryptoUtil.java
sajinieKavindya: +1

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
Nirothipan: If this is a conscious thing and we don't provide a default storage , is it correct to enable it by default .

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSInjectHandler.java
sajinieKavindya: changed to a boolean.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sajinieKavindya: Stopping the server if there are any DataSourceExceptions occurred while initializing the Transaction Count Handler component. Printing an error message otherwise.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sajinieKavindya: Changed the code to read the data source id only from deployment.toml.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
```diff
@@ -0,0 +1,142 @@
...
+                return takeRequestCuntOfTheMonth(synCtx, Integer.parseInt(yearParameter),
+                        Integer.parseInt(monthParameter));
+            } else if (StringUtils.isEmpty(yearParameter) && StringUtils.isEmpty(monthParameter)) {
+                Date date = new Date();
+                LocalDate localDate = date.toInstant().atZone(ZoneId.systemDefault()).toLocalDate();
+                return takeRequestCuntOfTheMonth(synCtx, localDate.getYear(), localDate.getMonthValue());
+            }
+        }
+        JSONObject response = Utils.createJsonError("Invalid Input", axisCtx, BAD_REQUEST);
+        axisCtx.removeProperty(org.wso2.micro.integrator.management.apis.Constants.NO_ENTITY_BODY);
```
Nirothipan, line 88: This should have been already included in setJsonPayload method by @sachithKay  .

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
```diff
@@ -0,0 +1,142 @@
...
+        Utils.setJsonPayLoad(axisCtx, response);
+        return true;
+    }
+
+    // Take aggregated request count for a given month.
+    private boolean takeRequestCuntOfTheMonth(MessageContext synCtx, int year, int month) {
```
sachithKay, line 94: typo in method signature

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
```diff
@@ -0,0 +1,142 @@
...
+        if (tryParseInt(yearParameter) != null && tryParseInt(monthParameter) != null) {
+            if (!StringUtils.isEmpty(yearParameter) && !StringUtils.isEmpty(monthParameter)) {
+                return takeRequestCuntOfTheMonth(synCtx, Integer.parseInt(yearParameter),
+                        Integer.parseInt(monthParameter));
+            } else if (StringUtils.isEmpty(yearParameter) && StringUtils.isEmpty(monthParameter)) {
+                Date date = new Date();
+                LocalDate localDate = date.toInstant().atZone(ZoneId.systemDefault()).toLocalDate();
+                return takeRequestCuntOfTheMonth(synCtx, localDate.getYear(), localDate.getMonthValue());
+            }
+        }
```
sdkottegoda, line 86: If the year and month is not present, I think we need to give the stats of the current month, WDYT?

sdkottegoda: along with the current year and the month

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Override
+    public boolean invoke(MessageContext synCtx) {
+
+        org.apache.axis2.context.MessageContext axisCtx = ((Axis2MessageContext) synCtx).getAxis2MessageContext();
+        String yearParameter = Utils.getQueryParameter(synCtx, "year");
+        String monthParameter = Utils.getQueryParameter(synCtx, "month");
+
+        if (tryParseInt(yearParameter) != null && tryParseInt(monthParameter) != null) {
+            if (!StringUtils.isEmpty(yearParameter) && !StringUtils.isEmpty(monthParameter)) {
```
sdkottegoda, line 78: can it be empty at this point? Wouldn't a numberformatException be thrown inside the method tryParseInt and a null be returned?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
```diff
@@ -0,0 +1,142 @@
...
+                return takeRequestCuntOfTheMonth(synCtx, Integer.parseInt(yearParameter),
+                        Integer.parseInt(monthParameter));
+            } else if (StringUtils.isEmpty(yearParameter) && StringUtils.isEmpty(monthParameter)) {
+                Date date = new Date();
+                LocalDate localDate = date.toInstant().atZone(ZoneId.systemDefault()).toLocalDate();
+                return takeRequestCuntOfTheMonth(synCtx, localDate.getYear(), localDate.getMonthValue());
+            }
+        }
+        JSONObject response = Utils.createJsonError("Invalid Input", axisCtx, BAD_REQUEST);
```
sachithKay, line 87: Shall we make the response more meaningful? Something like 'Input parameters are not valid'

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
sachithKay, line 88: This is redundant AFAIR. We are removing the NO_ENTITY_BODY in setJsonPayLoad method

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
sdkottegoda, line 87: Is BAD_REQUEST the correct status code? Wouldn't 403, forbidden suite better?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
```diff
@@ -0,0 +1,142 @@
...
+
+    // Take aggregated request count for a given month.
+    private boolean takeRequestCuntOfTheMonth(MessageContext synCtx, int year, int month) {
+
+        JSONObject response;
+        org.apache.axis2.context.MessageContext axisCtx = ((Axis2MessageContext) synCtx).getAxis2MessageContext();
+        axisCtx.setProperty(org.apache.axis2.Constants.Configuration.MESSAGE_TYPE, "application/json");
+        axisCtx.setProperty(Constants.Configuration.CONTENT_TYPE, "application/json");
+        // Sending a response body for a GET request
+        axisCtx.removeProperty(NO_ENTITY_BODY);
```
sdkottegoda, line 101: As nirothipan stated above Utils.setJsonPayLoad should fulfill these

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/store/connector/RDBMSConnector.java
```diff
@@ -78,35 +81,179 @@ private Connection getConnection() throws SQLException {
...
+                }
+            }
+        } finally {
+            closeAllConnections(dbConnection, rs, prepStmt);
+        }
+        return -1L;
+    }
 
+    // Add new entry to the transaction stat table
+    private void addNewRaw() {
```
sachithKay, line 129: should be corrected as addNewRow

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.wso2.micro.integrator.management.apis.Constants.BAD_REQUEST;
+import static org.wso2.micro.integrator.management.apis.Constants.INTERNAL_SERVER_ERROR;
+
+/**
+ * Resource for a retrieving aggregated request count.
+ * <p>
+ * Handles resources in the form "management/reqCount and management/reqCount?year=2020&month=5"
```
sdkottegoda, line 45: I think the comment is wrong. it should be management/transactions ?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
```diff
@@ -0,0 +1,142 @@
...
+            if (requestCount != -1L) {
+                response = new JSONObject(
+                        "{\"Year\" : " + year + ", \"Month\" : " + month + ", \"RequestCount\" : " + requestCount +
+                                "}");
+            } else {
+                response = new JSONObject("{\"error\" : \"Did not find stats for " + year + "/" + month + "\"}");
+            }
+        } else {
+            response =
+                    Utils.createJsonError("TransactionStore is not initialized", axisCtx, INTERNAL_SERVER_ERROR);
```
sdkottegoda, line 127: Is INTERNAL_SERVER_ERROR the correct status code? Wouldn't 403, forbidden suite better? because there's nothing wrong with the request

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
```diff
@@ -0,0 +1,142 @@
...
+        axisCtx.removeProperty(NO_ENTITY_BODY);
+
+        TransactionStore transactionStore = DataHolder.getInstance().getTransactionStore();
+        if (transactionStore != null) {
+            RDBMSConnector rdbmsConnector = transactionStore.getRdbmsConnector();
+            // Get request count of current Month
+            long requestCount;
+            try {
+                requestCount = rdbmsConnector.getTransactionCountOfMonth(year, month);
+            } catch (SQLException e) {
```
sdkottegoda, line 110: We shouldn't be catching an SQL exception here. This class shouldn't know anything about an sql implementation. Rather we should be calling,
if (transactionStore != null){
  long requestCount = transactionStore.getTransactionCountOfMonth(year,month);
}

the transactionStore will call rdbmsConnector.getTransactionCountOfMonth(year, month) upon that.

And the transaction store should be creating a TransactionCounterException upon receiving an SQL exception from it's rdbms connector

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
```diff
@@ -0,0 +1,142 @@
...
+                response =
+                        Utils.createJsonError("Error occurred while retrieving data from the database", axisCtx,
+                                INTERNAL_SERVER_ERROR);
+                Utils.setJsonPayLoad(axisCtx, response);
+                log.error("Error occurred while retrieving data from the database", e);
```
sdkottegoda, line 115: Utils.createJsonError does log the error aswell

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
```diff
@@ -0,0 +1,142 @@
...
+            }
+        } else {
+            response =
+                    Utils.createJsonError("TransactionStore is not initialized", axisCtx, INTERNAL_SERVER_ERROR);
+        }
+        Utils.setJsonPayLoad(axisCtx, response);
+        return true;
+    }
+
+    // try to parse user inputs to integer.
```
sdkottegoda, line 133: Let's add doc comments :)

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/store/connector/RDBMSConnector.java
```diff
@@ -78,35 +81,179 @@ private Connection getConnection() throws SQLException {
...
+            dbConnection = getConnection();
+            prepStmt = dbConnection.prepareStatement(TransactionQueryHelper.GET_TRAN_COUNT);
+            prepStmt.setString(1, getCurrentMonthAndYear());
+            prepStmt.setString(2, nodeId);
+            rs = prepStmt.executeQuery();
+            return rs.next();
+        } catch (SQLException e) {
+            LOG.error("Error occurred while reading transaction count table", e);
+        } finally {
+            closeAllConnections(dbConnection, rs, prepStmt);
```
sdkottegoda, line 173: Woudln't using try with resources be easier?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/store/connector/RDBMSConnector.java
```diff
@@ -78,35 +81,179 @@ private Connection getConnection() throws SQLException {
...
+            if (rs.next()) {
+                Object count = rs.getObject(1);
+                if (count != null) {
+                    return Long.parseLong(count.toString());
+                } else {
+                    return -1L;
+                }
+            }
+        } finally {
+            closeAllConnections(dbConnection, rs, prepStmt);
```
Nirothipan, line 123: Can't we simply use try with resources.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
GDLMadushanka: I wrote a method called tryParseInt to return null in the case of numberformatException

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
GDLMadushanka: +1 fixed

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
GDLMadushanka: +1 fixed

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
GDLMadushanka: +1 fixed

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/store/connector/RDBMSConnector.java
GDLMadushanka: +1 fixed

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
GDLMadushanka: +1 fixed

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
GDLMadushanka: +1 fixed

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
abeykoon: +1

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
GDLMadushanka: +1 fixed

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
GDLMadushanka: +1 fixed

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
GDLMadushanka: +1 fixed

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/store/connector/RDBMSConnector.java
GDLMadushanka: +1 fixed

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
```diff
@@ -0,0 +1,135 @@
...
+     *
+     * @param someText input string.
+     * @return parsed integer, null if failed.
+     */
+    private static Integer tryParseInt(String someText) {
+
+        try {
+            return Integer.parseInt(someText);
+        } catch (NumberFormatException ex) {
+            return null;
```
Nirothipan, line 132: Shall we print an error log here so that we can refer in server side.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
```diff
@@ -165,6 +173,24 @@ protected void activate(ComponentContext ctxt) {
 
                 synapseEnvironment.registerSynapseHandler(new SynapseExternalPropertyConfigurator());
                 synapseEnvironment.registerSynapseHandler(new ProxyLogHandler());
+
+                // Register internal transaction synapse handler
+                boolean transactionPropertyEnabled = TransactionCountHandlerComponent.isTransactionPropertyEnabled();
+                if (transactionPropertyEnabled) {
+                    TransactionCountHandler transactionCountHandler = new TransactionCountHandler();
+                    synapseEnvironment.registerSynapseHandler(transactionCountHandler);
```
Nirothipan, line 181: Seems transactionCountHandlerComponent.start doesn't need the handler , please correct me if I'm wrong. In that case shall we just register after starting it. Then we don't need to remove on error.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/store/TransactionStore.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Add transaction.
+     *
+     * @throws TransactionCounterException -
+     */
+    public void addTransaction() throws TransactionCounterException {
+        try {
+            this.rdbmsConnector.addTransaction();
+        } catch (SQLException e) {
```
Nirothipan, line 56: I think we should catch throw able here, if some other exception happens, the schedule task will be stopped thereafter. @sdkottegoda

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandlerComponent.java
```diff
@@ -0,0 +1,184 @@
...
+     * @param transactionStore - transactionStore instance.
+     */
+    private void scheduleTransactionWriterTask(TransactionStore transactionStore) {
+        txCountWriterTaskScheduler = Executors.newSingleThreadScheduledExecutor();
+        int scheduledPeriod = getTransactionCountScheduledPeriod();
+
+        txCountWriterTaskScheduler.scheduleAtFixedRate(() -> {
+            try {
+                transactionStore.addTransaction();
+            } catch (TransactionCounterException e) {
```
Nirothipan, line 108: I think we should catch throw able here, if some other exception happens, the schedule task will be stopped thereafter. @sdkottegoda

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.Set;
+
+import static org.wso2.micro.integrator.management.apis.Constants.*;
+
+/**
+ * Resource for a retrieving aggregated request count.
+ * <p>
+ * Handles resources in the form "management/transactions and management/transactions?year=2020&month=5"
+ */
+public class RequestCountResource extends ApiResource {
```
Nirothipan, line 43: Shouldn't this be extending ApiResourceAdapter .. + @sachithKay

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
sajinieKavindya: +1. fixed.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sajinieKavindya: catching a Throwable here to avoid any possible exceptions (other than Errors) that can cause this task to terminate.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandlerComponent.java
sajinieKavindya: +1

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
GDLMadushanka: yes, if both inputs are null we give the stats of the current month

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionHandlerComponent.java
sajinieKavindya: Disabling the Transaction count component by default. Hence, if the user wants to enable this feature, the required configurations will give them the notion that a valid data source should be configured priorly during the configuration time.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandlerComponent.java
```diff
@@ -44,15 +45,18 @@
 public class TransactionCountHandlerComponent {
 
     private static final Log LOG = LogFactory.getLog(TransactionCountHandlerComponent.class);
-    private static final int DEFAULT_SCHEDULED_PERIOD_INTERVAL = 60;
+    private static final int DEFAULT_UPDATE_INTERVAL = 1;
```
Nirothipan, line 48: Are we going to update every minute in a production server by default , isn't that too much . what about 15 minutes or something. Did we make this decision consciously ?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandlerComponent.java
```diff
@@ -69,27 +73,27 @@ public void start(DataSourceService dataSourceService) throws TransactionCounter
...
-                                 + DEFAULT_SCHEDULED_PERIOD_INTERVAL + "s)");
-                return DEFAULT_SCHEDULED_PERIOD_INTERVAL;
+                                 + DEFAULT_UPDATE_INTERVAL + "min)");
+                return DEFAULT_UPDATE_INTERVAL;
             }
         } else {
-            LOG.debug("Transaction count update period is set to default value (" + DEFAULT_SCHEDULED_PERIOD_INTERVAL
-                              + "s).");
-            return DEFAULT_SCHEDULED_PERIOD_INTERVAL;
+            LOG.debug("Transaction count update period is set to default value (" + DEFAULT_UPDATE_INTERVAL
```
Nirothipan, line 94: Shall we check if debug is enabled here

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandlerComponent.java
```diff
@@ -100,15 +104,17 @@ private int getTransactionCountScheduledPeriod() {
...
         txCountWriterTaskScheduler = Executors.newSingleThreadScheduledExecutor();
-        int scheduledPeriod = getTransactionCountScheduledPeriod();
+        int updateInterval = getTransactionCountUpdateInterval();
 
         txCountWriterTaskScheduler.scheduleAtFixedRate(() -> {
             try {
                 transactionStore.addTransaction();
-            } catch (TransactionCounterException e) {
-                LOG.debug("Could not persist the transaction count: ", e);
+                LOG.debug("Added the current transaction count: " + TransactionCountHandler.getTransactionCount()
```
Nirothipan, line 112: Same here

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/store/TransactionStore.java
sajinieKavindya: Disabling the Transaction count component by default. Hence, if the user wants to enable this feature, the required configurations will give them the notion that a valid data source should be configured priorly during the configuration time. As IMO, this will remove the need for a default data source being configured.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandlerComponent.java
sajinieKavindya: +1. fixed

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandlerComponent.java
sajinieKavindya: +1. fixed.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandlerComponent.java
sajinieKavindya: one minute was chosen to reduce the transaction count loss in a situation where a server shut down.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
```diff
@@ -522,4 +539,17 @@ public void setConfigAdminService(ConfigurationAdmin configAdminService) {
     public void unsetConfigAdminService(ConfigurationAdmin configAdminService) {
         log.debug("Unsetting ConfigurationAdmin Service");
     }
+
+    @Reference(name = "org.wso2.carbon.ndatasource",
```
sdkottegoda, line 543: This cannot have an impact on the startup time since we already have a dependency to TaskService which is already dependecnt on the ndatasource component. But shall we double check the startup time, just to be safe?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandler.java
```diff
@@ -0,0 +1,96 @@
...
+    public static long getTransactionCount() {
+        return transactionCount;
+    }
+
+    /**
+     * If the current transaction count has reached the Max Integer value, commit the current transaction to the
+     * database and reset the transaction count to zero.
+     */
+    private void commitAndResetTransactionCountIfReachedToMaxIntegerValue() {
+        if (transactionCount > 0 && (transactionCount + 1) < 0) {
```
sdkottegoda, line 76: I think it'll be more sensible if we just compare against Integer.MAX_VALUE, WDYT?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandlerComponent.java
sdkottegoda: One minute was chosen due to the following reasons. 
1. Assuming that we have a tps of 100, we will miss 100*60*5=30K requests if we configure the interval to 5 mins.
2. This is not affecting the direct messaging flow.
3. The performance impact of a simple db update per 1 min is negelgible,

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/store/connector/RDBMSConnector.java
```diff
@@ -0,0 +1,194 @@
...
+     * Get transaction count of a given year and month.
+     *
+     * @param year        Year.
+     * @param monthNumber Month.
+     * @return Aggregated transaction count if exists. Else (-1)
+     * @throws SQLException Error occurred while fetching data from the database.
+     */
+    public long getTransactionCountOfMonth(int year, int monthNumber) throws SQLException {
+
+        String dateString = year + "-" + monthNumber + "-01";
```
sdkottegoda, line 107: I hope this format is tested with all db types?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
sdkottegoda: Exactly. So, can yearParameter/monthParameter be empty without tryParseInt(yearParameter) returning null?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandler.java
sajinieKavindya: +1. fixed.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/ServiceBusInitializer.java
Nirothipan: Yes true , in that case do we need to specifically add this as a reference. Shall we neglect it as it is already implied.

sajinieKavindya: Checked. I could not notice a time difference in the startup time with this feature enabled.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/store/connector/RDBMSConnector.java
sajinieKavindya: @GDLMadushanka has tested it with all db types.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
sajinieKavindya: +1. fixed.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RequestCountResource.java
sachithKay: +1

sajinieKavindya: fixed.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandler.java
```diff
@@ -0,0 +1,96 @@
...
+    public static long getTransactionCount() {
+        return transactionCount;
+    }
+
+    /**
+     * If the current transaction count has reached the Max Integer value, commit the current transaction to the
+     * database and reset the transaction count to zero.
+     */
+    private void commitAndResetTransactionCountIfReachedToMaxIntegerValue() {
+        if (transactionCount == Integer.MAX_VALUE) {
```
Nirothipan, line 76: Shall we check this with some grace value and see if this is greater than that. Checking == might cause issues.
```suggestion
        if (transactionCount >= (Integer.MAX_VALUE - someGraceValue ) ) {
```

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandler.java
```diff
@@ -0,0 +1,96 @@
...
+        org.apache.axis2.context.MessageContext axis2MessageContext =
+                ((Axis2MessageContext) messageContext).getAxis2MessageContext();
+        Object transactionProperty = axis2MessageContext.getProperty(BaseConstants.INTERNAL_TRANSACTION_COUNTED);
+        // increment the transaction count by 1 if the INTERNAL_TRANSACTION_COUNTED message property has not been set
+        // or if the value of the property is set to false.
+        if (!(transactionProperty instanceof Boolean && (Boolean) transactionProperty)) {
+            axis2MessageContext.setProperty(BaseConstants.INTERNAL_TRANSACTION_COUNTED, true);
+            // commit the current transaction count to database and set the transaction count to zero if it has
+            // reached to Interger.MAX_VALUE
+            commitAndResetTransactionCountIfReachedToMaxIntegerValue();
```
Nirothipan, line 46: Shall we execute this periodically , with some lesser period. True that it won't be 100% accurate but executing this ( checking int max value ) for each and every request might affect performance in other hand . @sdkottegoda

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandler.java
sdkottegoda: +1

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandler.java
sdkottegoda: Checking against int max will not cause a considerable degradation. But, we can avoid it by making the transaction handling part asynchronous.

sdkottegoda: However, I just realized that this method is one that will definitely be called concurrently and we need to have some synchronization in updating the counter.

So I think, we need to notify an async Counter thread which will update the transaction count in a synchronized method. @sajinieKavindya

sajinieKavindya: I think the performance effect this simple check can cause if negligible when compared to all the other processings done through the message flow of a request. And if we run this periodically, there is a chance that the transaction count gets exceeded the int max value by that time this value is checked, which will cause some issues afterward. But if we are going to check the count with a grace value as suggested in https://github.com/wso2/micro-integrator/pull/1537#discussion_r425904061, maybe we can eliminate that issue. But it again depends on how often we are going to execute this check. (Therefore, this update interval and the grace value needs to be chosen wisely in that case). @sdkottegoda

Nirothipan: True it is of negligible performance impact. But IMO better to avoid them wherever possible. However moving to an async thread will sort all these issues I think. Also shall we verify this with concurrent requests to ensure the level of correctness. ( To test concurrent requests you can simply do in a  single command with Apache benchmark )

sajinieKavindya: This issue will be addressed separately here https://github.com/wso2/micro-integrator/issues/1558

PR #1550, opened 2020-05-14
Add db scripts for transaction count feature.

GDLMadushanka:


distribution/src/dbscripts/mysql_transaction_count.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS CURRENT_STATS (
+  TIME_STAMP DATE NOT NULL,
+  NODE_ID VARCHAR(40) NOT NULL,
+  TRANSACTION_COUNT INT NOT NULL,
+  TRANSACTION_COUNT_ENCRYPTED TEXT NOT NULL
```
sachithKay, line 5: shall we make this into VARCHAR? Better if we can keep it in a dynamically allocated type

distribution/src/dbscripts/mssql_transaction_count.sql
```diff
@@ -0,0 +1,7 @@
+IF NOT  EXISTS (SELECT * FROM SYS.OBJECTS WHERE OBJECT_ID = OBJECT_ID(N'[DBO].[CURRENT_STATS]') AND TYPE IN (N'U'))
+CREATE TABLE  CURRENT_STATS (
+  TIME_STAMP DATE NOT NULL,
+  NODE_ID VARCHAR (40) NOT NULL,
+  TRANSACTION_COUNT INT NOT NULL,
+  TRANSACTION_COUNT_ENCRYPTED TEXT NOT NULL
```
sachithKay, line 6: dynamic allocation type would be better

GDLMadushanka: I think by dynamic allocation you mean dynamic memory allocation. If so, text type is also dynamically allocated. It's only deprecated in MySQL. Other DB types continue to support this type. I will fix that in MySql script.

sachithKay: seems text is to be deprecated by SQL server too. https://docs.microsoft.com/en-us/sql/database-engine/deprecated-database-engine-features-in-sql-server-2017?view=sql-server-ver15

Their recommendation is to use varchar(max). Isn't it better it we use VARCHAR consistently over our scripts? WDYT (If other databases have it)?

PR #1554, opened 2020-05-15
Log number of occurrences for testNestedForEachMediatorWithIterate

sachithKay:

### DIFF
--- a/integration/mediation-tests/tests-mediator-1/src/test/java/org/wso2/carbon/esb/mediator/test/foreach/NestedForEachTestCase.java
+++ b/integration/mediation-tests/tests-mediator-1/src/test/java/org/wso2/carbon/esb/mediator/test/foreach/NestedForEachTestCase.java
@@ -93,7 +93,6 @@ public class NestedForEachTestCase extends ESBIntegrationTest {
 
     @Test(groups = "wso2.esb", description = "Transforming a Message Using a Nested ForEach Construct with Iterate/Aggregate Sending Payload to backend")
     public void testNestedForEachMediatorWithIterate() throws Exception {
-        carbonLogReader.clearLogs();
         String response = client.send(getProxyServiceURLHttp("nested_foreach_iterate"), createMultipleSymbolPayLoad(10), "urn:getQuote");
         Assert.assertNotNull(response);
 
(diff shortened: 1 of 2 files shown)


integration/mediation-tests/tests-mediator-1/src/test/java/org/wso2/carbon/esb/mediator/test/foreach/NestedForEachTestCase.java
```diff
@@ -111,9 +110,11 @@ public void testNestedForEachMediatorWithIterate() throws Exception {
             }
         }
         Assert.assertTrue(carbonLogReader.checkForLog("foreach = outer", DEFAULT_TIMEOUT, 10),
-                "Count of messages entered outer ForEach scope is incorrect");
+                "Count of messages entered outer ForEach scope is incorrect. " +
+                        "Found " + carbonLogReader.getNumberOfOccurencesForLog("foreach = outer") + " occurrences");
```
sdkottegoda, line 114: Shall we do this inside the method, carbonLogReader.checkForLog(...)? The reason is, this is not the only test case that is failing this way and we will have to add the message to all the tests unless we add it somewhere common.

sachithKay: Added a Log to checkForLog method as well.

integration/tests-common/integration-test-utils/src/main/java/org/wso2/esb/integration/common/utils/CarbonLogReader.java
```diff
@@ -178,6 +182,7 @@ public boolean checkForLog(String expected, int timeout, int numberofOccurrences
             }
             TimeUnit.SECONDS.sleep(1);
         }
+        LOG.warn("Found "+ this.getNumberOfOccurencesForLog(expected) +" occurrences while expecting " + numberofOccurrences);
```
Nirothipan, line 185: In this case , I thought it would be better to print the whole content of the logs in within an identifiable string like 
===========
Whole logs
===========

so that we can manually see whether that content is there in the logs checked. Coz we could see them in logs downloaded but the missing part is not being  sure whether it is collected by log tailer.

sachithKay: added the log to the same warn

Issue #1558, opened 2020-05-18
Address concurrency issues in Transaction Count Handler


sajinieKavindya:
**Description:**
handleRequestInFlow method in the TransactionCountHandler will definitely be called concurrently. Hence, we need to have some synchronization in updating the counter.

**Affected Product Version:**
1.2.0

PR #1560, opened 2020-05-19
Fix concurrency issues in Transaction Count Handler

sajinieKavindya:
## Purpose
The workload of the transaction counting part should be handled asynchronously to have the least performance impact on the main request flow due to this functionality. In order to achieve this, this PR has moved the transaction count handling part to a separate thread to make it asynchronous.

Also, the handleRequestInFlow method in the TransactionCountHandler will definitely be called concurrently and therefore, the variable transaction count will be modified by different threads. Hence, this PR has made the counter updating functionality synchronized. 

Fix: https://github.com/wso2/micro-integrator/issues/1558

### DIFF
--- a/components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandler.java
+++ b/components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandler.java
@@ -27,24 +27,27 @@ import org.apache.synapse.core.axis2.Axis2MessageContext;
 import org.wso2.micro.integrator.initializer.handler.DataHolder;
 import org.wso2.micro.integrator.initializer.handler.transaction.store.TransactionStore;
 
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.RejectedExecutionException;
+
 public class TransactionCountHandler extends AbstractSynapseHandler {
 
     private static final Log LOG = LogFactory.getLog(TransactionCountHandler.class);
-    private static int transactionCount = 0;
+    private static final double MAX_TRANSACTION_COUNT = Integer.MAX_VALUE * 0.9;
+    private static volatile int transactionCount = 0;
+    private ExecutorService transactionCountExecutor;
+    private final Object lock = new Object();
+
+    public TransactionCountHandler(ExecutorService executor) {
+        this.transactionCountExecutor = executor;
+    }
 
     @Override
     public boolean handleRequestInFlow(MessageContext messageContext) {
-        org.apache.axis2.context.MessageContext axis2MessageContext =
-                ((Axis2MessageContext) messageContext).getAxis2MessageContext();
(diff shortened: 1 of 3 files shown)


Nirothipan:
Were you able to verify this with concurrent requests and some service chaining , what was the accuracy .
Also are we planning to include integration tests for these .

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandler.java
```diff
@@ -68,29 +58,47 @@ public static long getTransactionCount() {
...
+        Object transactionProperty = axis2MessageContext.getProperty(BaseConstants.INTERNAL_TRANSACTION_COUNTED);
+        // increment the transaction count by 1 if the INTERNAL_TRANSACTION_COUNTED message property has not been set
+        // or if the value of the property is set to false.
+        if (!(transactionProperty instanceof Boolean && (Boolean) transactionProperty)) {
+            axis2MessageContext.setProperty(BaseConstants.INTERNAL_TRANSACTION_COUNTED, true);
+            // commit the current transaction count to database and set the transaction count to zero if it has
+            // reached to Interger.MAX_VALUE
+            commitAndResetTransactionCountIfReachedToMaxIntegerValue();
+            incrementTransactionCount();
+        }
```
Nirothipan, line 73: shall we add some debug logs for this method , also in else part.

sajinieKavindya: Done.

sajinieKavindya:
> Were you able to verify this with concurrent requests , what was the accuracy .
Did performance tests with 6715.8 requests per second for an hour. Accuracy was 100%. 
> Also are we planning to include integration tests for these .
Yes.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionCountHandler.java
```diff
@@ -69,28 +71,56 @@ public static long getTransactionCount() {
...
+    }
+
+    /**
+     * If the current transaction count has reached or exceeded the Max_Integer_value * 0.9, commit the current
+     * transaction to the database, set a new node id, and reset the transaction count to zero.
      */
     private void commitAndResetTransactionCountIfReachedToMaxIntegerValue() {
-        if (transactionCount == Integer.MAX_VALUE) {
+        if (transactionCount >= Integer.MAX_VALUE * 0.9) {
```
sachithKay, line 108: Shall we assign this (Integer.MAX_VALUE * 0.9) to a const and use it here

sajinieKavindya: Done.

PR #1562, opened 2020-05-19
Refactor Error responses for artifact suspension

sachithKay:

### DIFF
--- a/components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/EndpointResource.java
+++ b/components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/EndpointResource.java
@@ -195,13 +195,12 @@ public class EndpointResource implements MiApiResource {
                 ep.getContext().switchOn();
                 jsonResponse.put(Constants.MESSAGE_JSON_ATTRIBUTE, endpointName + " is switched On");
             } else {
-                LOG.warn("Specified Status: " + status + " is not valid");
-                jsonResponse.put(Constants.MESSAGE_JSON_ATTRIBUTE, "Provided state is not valid");
+                jsonResponse = Utils.createJsonError("Provided state is not valid", axis2MessageContext, Constants.BAD_REQUEST);
             }
             Utils.setJsonPayLoad(axis2MessageContext, jsonResponse);
         } else {
-            LOG.warn("Specified Endpoint: " + endpointName + " does not exist");
-            Utils.setJsonPayLoad(axis2MessageContext, Utils.createJsonErrorObject("Endpoint does not exist"));
+            Utils.setJsonPayLoad(axis2MessageContext,  Utils.createJsonError("Endpoint does not exist",
+                    axis2MessageContext, Constants.NOT_FOUND));
         }
     }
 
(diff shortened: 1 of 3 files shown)


components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/EndpointResource.java
```diff
@@ -196,12 +196,12 @@ private void changeEndpointStatus(org.apache.axis2.context.MessageContext axis2M
...
                 LOG.warn("Specified Status: " + status + " is not valid");
-                jsonResponse.put(Constants.MESSAGE_JSON_ATTRIBUTE, "Provided state is not valid");
+                jsonResponse = Utils.createJsonError("Provided state is not valid", axis2MessageContext, "400");
             }
             Utils.setJsonPayLoad(axis2MessageContext, jsonResponse);
         } else {
-            LOG.warn("Specified Endpoint: " + endpointName + " does not exist");
-            Utils.setJsonPayLoad(axis2MessageContext, Utils.createJsonErrorObject("Endpoint does not exist"));
+            Utils.setJsonPayLoad(axis2MessageContext,  Utils.createJsonError("Endpoint does not exist",
+                    axis2MessageContext, "403"));
```
Vathsan, line 204: Wouldn't it be better if we set 404 instead of 403 here? WDYT?

sachithKay: 404 is used when Request-URI is unavailable.
Ideally web resource is available right? it's only the endpoint the user is trying to deactivate is missing. WDYT?

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/MessageProcessorResource.java
```diff
@@ -224,12 +224,12 @@ private void changeProcessorStatus(MessageContext messageContext, JsonObject jso
                 messageProcessor.activate();
                 jsonResponse.put(Constants.MESSAGE_JSON_ATTRIBUTE, processorName + " : is activated");
             } else {
-                jsonResponse.put(Constants.MESSAGE_JSON_ATTRIBUTE, "Provided state is not valid");
+                jsonResponse = Utils.createJsonError("Provided state is not valid", axis2MessageContext, "400");
```
sdkottegoda, line 227: Let's re-use the constants defined for these error codes

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/EndpointResource.java
sdkottegoda: Ideally the request accessed here (the endpoint) should have been in the request path and the request will match the spec of 404 in that case. so I think 404 is more applicable. 403, is also fine since we are giving the reason for forbidding the request.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/MessageProcessorResource.java
sachithKay: +1. Done

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/EndpointResource.java
sachithKay: Changed the non-existing resources to 404 SC

PR #1586, opened 2020-05-25
Add Public Key based Encryption to Transaction Count Handler Component

sajinieKavindya:
## Purpose
Before this PR, the encryption was done using the public key certificate inside the default WSO2 Truststore. This PR has changed the encryption functionality to encrypt using a new public key, which is generated for the use of this Transaction Count Handler component.


components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionConstants.java
```diff
@@ -18,18 +18,16 @@
...
-    public static final String DEFAULT_SECURITY_RESOURCE_DIR_PATH = System.getProperty(ServerConstants.CARBON_HOME)
-            + File.separator + "repository" + File.separator + "resources" + File.separator + "security";
-    public static final String TRUSTSTORE_FILE = "wso2carbon.jks";
-
-    public static final String KEY_TYPE = "JKS";
-    public static final String KEY_ALIAS = "wso2carbon";
-    public static final String KEYSTORE_PASSWORD = "wso2carbon";
     public static final String ENCRYPTION_ALGORITHM = "RSA";
+
+    public static final String ENCRYPTION_PUB_KEY = "MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAuXUnLzRfEokhRQT62dnh\n"
```
Nirothipan, line 26: how was this generated. will this change if we change default certs

milindaperera:
We cannot hard code the public key. We should loaded from the certificate file at startup. Mainly because we may be need to do change in the public key, with time progresses. In such cases we have to do code level changes. That is not acceptable for a resource. Mainly because it's not a constant in long run.

So we have two options:
1. At least load from the jar resource 
2. Load from file system (gives more freedom for customizations)

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/handler/transaction/TransactionConstants.java
sajinieKavindya: How was this generated?
- The key pair was newly generated using OpenSSL for the use of this transaction count handler component

Will this change if we change default certs?
- No, because the public key is not embedded into any certificate (simply because we are not intending to do any secure communication or any such activities using this public key).

Issue #1602, opened 2020-05-29
Enable hot deployment by default in zip distribution


Nirothipan:
**Description:**
$subject to cater the vm based deployments since zip packaging is oriented towards it.

PR #1606, opened 2020-06-01
Enable hot deployment by default

Nirothipan:
Fixes https://github.com/wso2/micro-integrator/issues/1602

Doc PR https://github.com/wso2/docs-ei/pull/1612


integration/mediation-tests/tests-other/src/test/resources/artifacts/ESB/hotdeployment/deployment.toml
```diff
@@ -1,6 +1,6 @@
 [server]
 hostname = "localhost"
-hot_deployment = "true"
+hot_deployment = "false"
```
sajinieKavindya, line 3: Shall we use the boolean value instead of a String here as well?

Nirothipan: In other places this is given as boolean, I left this this way just to make sure that both string and boolean value works for hot deployment. wdyt.

Issue #1601, opened 2020-05-29
Can not resolve secrets in Datasources


malakaganga:
**Description:**
Can not resolve secrets in Datasources

**Steps to reproduce:**
Add the following data source in deployment.toml

[[datasource]]
id = "WSO2_COORDINATION_DB"
url= "jdbc:mysql://localhost:3306/clusterdb"
username="root"
password="$secret{db_pass}"
driver="com.mysql.jdbc.Driver"

And when loading the server following exception throws

```Error in initializing Tasks component: Error in connecting to the datasource to create tables for clustering org.wso2.micro.integrator.coordination.exception.ClusterCoordinationException: Error in connecting to the datasource to create tables for clustering
	at org.wso2.micro.integrator.coordination.RDBMSCommunicationBusContextImpl.<init>(RDBMSCommunicationBusContextImpl.java:69)
	at org.wso2.micro.integrator.coordination.RDBMSCoordinationStrategy.<init>(RDBMSCoordinationStrategy.java:134)
	at org.wso2.micro.integrator.coordination.ClusterCoordinator.<init>(ClusterCoordinator.java:40)
	at org.wso2.micro.integrator.ntask.core.internal.TasksDSComponent.activate(TasksDSComponent.java:118)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.eclipse.equinox.internal.ds.model.ServiceComponent.activate(ServiceComponent.java:260)
	at org.eclipse.equinox.internal.ds.model.ServiceComponentProp.activate(ServiceComponentProp.java:146)
	at org.eclipse.equinox.internal.ds.model.ServiceComponentProp.build(ServiceComponentProp.java:345)
	at org.eclipse.equinox.internal.ds.InstanceProcess.buildComponent(InstanceProcess.java:620)
	at org.eclipse.equinox.internal.ds.InstanceProcess.buildComponents(InstanceProcess.java:197)
	at org.eclipse.equinox.internal.ds.Resolver.buildNewlySatisfied(Resolver.java:473)
[...]

PR #1611, opened 2020-06-02
Fix https://github.com/wso2/micro-integrator/issues/1601

malakaganga:
## Purpose
Give capability to secure passwords in master-datasource.xml

Fixes: https://github.com/wso2/micro-integrator/issues/1601

### DIFF
--- a/components/org.wso2.micro.integrator.ndatasource.core/src/main/java/org/wso2/micro/integrator/ndatasource/core/DataSourceManager.java
+++ b/components/org.wso2.micro.integrator.ndatasource.core/src/main/java/org/wso2/micro/integrator/ndatasource/core/DataSourceManager.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import javax.xml.bind.JAXBContext;
 
+import org.apache.axiom.om.OMElement;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.w3c.dom.Document;
@@ -175,11 +176,11 @@ public class DataSourceManager {
 
 	private void initSystemDataSource(File sysDSFile) throws DataSourceException {
 		try {
-		    JAXBContext ctx = JAXBContext.newInstance(SystemDataSourcesConfiguration.class);
-            Document doc = DataSourceUtils.convertToDocument(sysDSFile);
-            DataSourceUtils.secureResolveDocument(doc, true);
-		    SystemDataSourcesConfiguration sysDS = (SystemDataSourcesConfiguration) ctx.createUnmarshaller().
-		    		unmarshal(doc);
+			JAXBContext ctx = JAXBContext.newInstance(SystemDataSourcesConfiguration.class);
+			OMElement doc = DataSourceUtils.convertToOMElement(sysDSFile);
+			DataSourceUtils.secureResolveOMElement(doc);
+			SystemDataSourcesConfiguration sysDS = (SystemDataSourcesConfiguration) ctx.createUnmarshaller().
+					unmarshal(doc.getXMLStreamReader());
 		    this.addDataSourceProviders(sysDS.getProviders());
 		    DataSourceRepository dsRepo = this.getDataSourceRepository(
 				    Constants.SUPER_TENANT_ID);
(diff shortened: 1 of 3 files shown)


components/org.wso2.micro.integrator.ndatasource.core/src/main/java/org/wso2/micro/integrator/ndatasource/core/utils/DataSourceUtils.java
```diff
@@ -244,6 +251,46 @@ public static void secureResolveDocument(Document doc, boolean checkSecureVault)
 		}
     }
 
+	public static void secureResolveOMElement(OMElement doc, boolean checkSecureVault)
+			throws DataSourceException {
+
```
sajinieKavindya, line 256: It seems like the boolean checkSecureVault has no usage inside this method.

components/org.wso2.micro.integrator.ndatasource.core/src/main/java/org/wso2/micro/integrator/ndatasource/core/DataSourceManager.java
```diff
@@ -175,11 +176,11 @@ public void initSystemDataSources() throws DataSourceException {
 
 	private void initSystemDataSource(File sysDSFile) throws DataSourceException {
 		try {
-		    JAXBContext ctx = JAXBContext.newInstance(SystemDataSourcesConfiguration.class);
-            Document doc = DataSourceUtils.convertToDocument(sysDSFile);
-            DataSourceUtils.secureResolveDocument(doc, true);
-		    SystemDataSourcesConfiguration sysDS = (SystemDataSourcesConfiguration) ctx.createUnmarshaller().
```
sajinieKavindya, line 181: DataSourceUtils.secureResolveDocument method is used in DataSourceCappDeployer class. Maybe that also needs to be replaced with DataSourceUtils.secureResolveOMElement method if this method is not functioning properly (But it would be a separate issue). WDYT?

malakaganga: Yes Deploying a Datasource with Capp will not work if the password is secured but I couldn't find a doc regarding securing passwords when deploying a Datasource through capps if we are doing that we may need to add tooling support and docs as well that's why I have left it untouched.

components/org.wso2.micro.integrator.ndatasource.core/src/main/java/org/wso2/micro/integrator/ndatasource/core/utils/DataSourceUtils.java
```diff
@@ -244,6 +251,46 @@ public static void secureResolveDocument(Document doc, boolean checkSecureVault)
...
+	}
+
+	public static OMElement convertToOMElement(File file) throws DataSourceException {
+
+		try {
+			OMElement documentElement = new StAXOMBuilder(new FileInputStream(file)).getDocumentElement();
+			return documentElement;
+		} catch (Exception e) {
+			throw new DataSourceException("Error in creating an XML document from file: " +
+					e.getMessage(), e);
```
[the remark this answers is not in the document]
malakaganga: It has been added for the sake of finishing the sentence.

components/org.wso2.micro.integrator.ndatasource.core/src/main/java/org/wso2/micro/integrator/ndatasource/core/utils/DataSourceUtils.java
malakaganga: Fixed

components/org.wso2.micro.integrator.ndatasource.core/src/main/java/org/wso2/micro/integrator/ndatasource/core/utils/DataSourceUtils.java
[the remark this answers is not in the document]
sajinieKavindya: Can't we just end the sentence with the file name (file.getAbsolutePath()) and let the detailed message e.message() be there in the exception itself. I think it is not necessary to log this separately whenever we are passing the exception along with the DataSourceException object because it will anyway gets printed with the exception. WDYT?

PR #1627, opened 2020-06-04
Upgrade xerces dependency version

sajinieKavindya:
## Purpose
Modified MI to use xerces wso2 orbit dependency and upgraded the version to 2.12.0.wso2v1. This orbit bundle uses xerces:xercesImpl:2.12.0.


Nirothipan:
Is this being used in cipher tool or sthng. Please verify whether they are working properly with new versions

sajinieKavindya:
> Is this being used in cipher tool or sthng. Please verify whether they are working properly with new versions

xerces is not used in cipher-tool.

PR #1631, opened 2020-06-04
Upgrade jackson dependencies

sajinieKavindya:
## Purpose
Upgrade jackson dependencies to version 2.10.3.

### DIFF
--- a/components/mediation/data-publishers/org.wso2.micro.integrator.analytics.messageflow.data.publisher/pom.xml
+++ b/components/mediation/data-publishers/org.wso2.micro.integrator.analytics.messageflow.data.publisher/pom.xml
@@ -53,11 +53,11 @@
             <artifactId>org.wso2.micro.integrator.analytics.data.publisher.util</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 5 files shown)


distribution/src/assembly/bin.xml
```diff
@@ -608,9 +608,9 @@
                 <include>com.google.guava:guava:jar</include>
                 <include>org.jsoup:jsoup:jar</include>
                 <include>com.google.re2j:re2j:jar</include>
-                <include>com.fasterxml.jackson.core:jackson-core:jar</include>
-                <include>com.fasterxml.jackson.core:jackson-databind:jar</include>
-                <include>com.fasterxml.jackson.core:jackson-annotations:jar</include>
+                <include>org.wso2.orbit.com.fasterxml.jackson.core:jackson-core:jar</include>
+                <include>org.wso2.orbit.com.fasterxml.jackson.core:jackson-databind:jar</include>
+                <include>org.wso2.orbit.com.fasterxml.jackson.core:jackson-annotations:jar</include>
```
Nirothipan, line 613: Shall we better leave direct bundles here

PR #1634, opened 2020-06-06
Enable user store db tests

Nirothipan:
$subject


distribution/src/dbscripts/oracle/oracle_user.sql
```diff
@@ -459,8 +459,7 @@ CREATE TABLE UM_SYSTEM_ROLE(
             UM_ID INTEGER NOT NULL,
             UM_ROLE_NAME VARCHAR(255) NOT NULL,
             UM_TENANT_ID INTEGER DEFAULT 0,
-            PRIMARY KEY (UM_ID, UM_TENANT_ID),
-            UNIQUE(UM_ROLE_NAME,UM_TENANT_ID)
```
sdkottegoda, line 463: Removing this, brings down the unique constraint, does it not?

Nirothipan: The same index is added in line 468
CREATE INDEX SYSTEM_ROLE_IND_BY_RN_TI ON UM_SYSTEM_ROLE(UM_ROLE_NAME, UM_TENANT_ID)
Moreover, I couldn't find the same in the script from the kernel.

PR #1638, opened 2020-06-08
Enable ManagementAPI by default in MI

sajinieKavindya:
## Purpose
This PR will enable the management API by default in MI.


Nirothipan:
Shall we add to wrapper.conf and also remove or make this as info https://github.com/wso2/micro-integrator/blob/master/components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/ManagementInternalApi.java#L65

Nirothipan:
Also we need the relevant docs also to be updated

PR #1648, opened 2020-06-09
Prevent defining illegal registry key

arunans23:
Prevent defining illegal registry key

### DIFF
--- a/components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
+++ b/components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
@@ -515,7 +515,8 @@ public class MicroIntegratorRegistry extends AbstractRegistry {
             String parent = getParentPath(targetPath, isDirectory);
             try {
                 if (isDirectory) {
-                    if (!new File(parent).exists() && !new File(parent).mkdirs()) {
+                    File parentFile = new File(parent);
+                    if (!parentFile.exists() && !parentFile.mkdirs()) {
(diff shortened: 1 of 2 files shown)


components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
```diff
@@ -861,14 +862,22 @@ private String resolveRegistryURI(String key) {
...
-
             resolvedPath = registryRoot + resourcePath;
-        }
 
+            //Test whether registry key has any illegel access
+            File resolvedPathFile = new File(resolvedPath);
+            File registryRootFile = new File(registryRoot);
+            try {
+                if (!resolvedPathFile.getCanonicalPath().startsWith(registryRootFile.getCanonicalPath())) {
+                    handleException("The registry key defined is illegal");
```
sdkottegoda, line 872: Should we log the resolved path? and the registry root path?

arunans23: In this case, we do not need to. This method will be called several times for various resources. Maybe we can add a debug log. WDYT?

Nirothipan: +1 to print registry root path but I'm not sure we can print the illegal path in the logs.

arunans23: Logged the registry key

PR #1652, opened 2020-06-10
Remove unused deprecated method from UserCoreUtil

arunans23:
Remove unused deprecated method `getPasswordToStore()` from UserCoreUtil


sdkottegoda:
Why are we removing this? Is it just because it's not used? Or is there some issue with this code? If that is the case, I suppose you checked with the non-deprecated method public static byte[] getPasswordToStore() for the same issue?

arunans23:
> Why are we removing this? Is it just because it's not used? Or is there some issue with this code? If that is the case, I suppose you checked with the non-deprecated method public static byte[] getPasswordToStore() for the same issue?

Both. It is not used anywhere in MI. This code is moved from kernel. It has no use here. Also there are some issues with this code. non-deprecated method does not have the issue.

PR #1687, opened 2020-06-15
Create git action workflow for windows 

arunans23:
Create git action workflow for windows to run full Integration tests in a split manner (Due to the excess startup time of the server)

### DIFF
--- a/.github/workflows/cluster-tests.yml
+++ b/.github/workflows/cluster-tests.yml
@@ -1,4 +1,4 @@
-name: Cluster tests ( openjdk11 & mysql )
+name: Cluster tests (OpenJDK11 & MySQL)
 
 on: [push, pull_request]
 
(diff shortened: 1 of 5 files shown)


.github/workflows/cluster-tests.yml
```diff
@@ -1,4 +1,4 @@
-name: Cluster tests ( openjdk11 & mysql )
+name: Cluster tests (Ubuntu, OpenJDK11 & MySQL)
```
Nirothipan, line 1: Ubuntu will already be there "Cluster tests (Ubuntu, OpenJDK11 & MySQL) / ubuntu (pull_request) " 
It is redundant.

sdkottegoda:
Is it possible to name them with the module names rather than just windows-1, windows-2?

PR #1699, opened 2020-06-17
Remove PrometheusStatisticsTest Test Case

Nipunich:
Remove PrometheusStatisticsTest Test Case because it's changed in the with the new observability implementation.

### DIFF
--- a/integration/mediation-tests/tests-other/src/test/resources/testng.xml
+++ b/integration/mediation-tests/tests-other/src/test/resources/testng.xml
@@ -118,6 +118,11 @@
                     <exclude name=".*"/>
                 </methods>
             </class>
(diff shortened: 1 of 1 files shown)


integration/mediation-tests/tests-other/src/test/resources/testng.xml
```diff
@@ -118,6 +118,11 @@
                     <exclude name=".*"/>
                 </methods>
             </class>
+            <class name="org.wso2.carbon.esb.statistics.PrometheusStatisticsTest">
+                <methods>
+                    <exclude name=".*"/>
+                </methods>
```
Nirothipan, line 124: Please remove the test case and not skip this. @arunans23 @Nipunich

Issue #1735, opened 2020-06-24
Registry resources are not getting removed upon the capp undeployment


chanikag:
**Description:**
The registry resources are remaining even the capp is undeployed

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Related Issues:**

PR #1736, opened 2020-06-24
Fix MI-1735: Registry resources are not getting removed upon the capp undeployment

chanikag:
This is to add the logic to remove registry resources when the capp is getting undeployed

To fix : https://github.com/wso2/micro-integrator/issues/1735

### DIFF
--- a/components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
+++ b/components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
@@ -838,7 +838,7 @@ public class MicroIntegratorRegistry extends AbstractRegistry {
         String metadataDirPath = parent.getPath() + File.separator + METADATA_DIR_NAME;
         File metadataDir = new File(metadataDirPath);
         if (!metadataDir.exists()) {
-            handleException("Unable to delete metadata directory: " + metadataDir.getPath());
+            return;
         }
         File metadataFile = new File(metadataDir, resourceFileName + METADATA_FILE_SUFFIX);
         deleteFile(metadataFile);
--- a/components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
+++ b/components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
@@ -837,7 +837,7 @@ public class MicroIntegratorRegistry extends AbstractRegistry {
 
         String metadataDirPath = parent.getPath() + File.separator + METADATA_DIR_NAME;
         File metadataDir = new File(metadataDirPath);
-        if (!metadataDir.exists() && !metadataDir.mkdirs()) {
+        if (!metadataDir.exists()) {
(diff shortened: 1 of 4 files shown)


components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/RegistryHelper.java
```diff
@@ -69,4 +69,18 @@ public static String getSystemDependentPath(String path) {
...
+     * Check whether the directory is empty
+     *
+     * @param directory  directory path
+     */
+    public static boolean isDirectoryEmpty (String directory) {
+        File file = new File(directory);
+        if (file.isDirectory()) {
+            return file.list().length == 0;
+        } else {
+            return false;
```
sdkottegoda, line 82: Should we return an error since we are expecting 'directory' to be a directory?

chanikag: The method is to check whether the directory is empty. We don't want to fail the flow, if we are checking empty when it is not a directory, for an example if it is a file. It just says false. WDYT?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/deployment/synapse/deployer/FileRegistryResourceDeployer.java
```diff
@@ -179,6 +205,29 @@ private void writeArtifactToRegistry(RegistryConfig registryConfig){
...
+
+        // get resources
+        List<RegistryConfig.Resourse> resources = registryConfig.getResources();
+        for (RegistryConfig.Resourse resource : resources) {
+            String filePath = registryConfig.getExtractedPath() + File.separator + AppDeployerConstants.RESOURCES_DIR
+                    + File.separator + resource.getFileName();
+            // check whether the file exists
+            File file = new File(filePath);
+            if (!file.exists()) {
+                log.error("Specified file to be removed as a resource is " + "not found at : " + filePath);
```
Nirothipan, line 223: when clustered , another node would have deleted it already. should we simply ignore this rather than printing the error msg or debug ?

chanikag: +1.

components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
```diff
@@ -473,7 +475,19 @@ public OMNode lookupFormat(String key) {
...
+                File parentFile = new File(new URI(getParentPath(targetPath, false)));
+                String fileName = getResourceName(targetPath);
+                removeResource(path);
+                deleteMetadata(parentFile, fileName);
+                if (RegistryHelper.isDirectoryEmpty(parentFile.getPath()) && !CONFIG_DIRECTORY_NAME.equals
+                        (parentFile.getName()) && !GOVERNANCE_DIRECTORY_NAME.equals(parentFile.getName())) {
+                    deleteDirectory(parentFile);
+                }
+            } catch (URISyntaxException e) {
+                handleException("Error while deleting the registry path " + path);
```
Nirothipan, line 489: shall we pass exception as well.

PR #1737, opened 2020-06-24
Fix EI-5133: Introduce a property to hide current params in the DSFault msg

chanikag:
## Purpose

This is to fix wso2/product-ei#5133

If the current params should be hidden in the log file and then in the wsdl, the following system property should be enabled.
dss.disable.current.params=true

### DIFF
--- a/components/data/data-services/org.wso2.micro.integrator.dataservices.core/src/main/java/org/wso2/micro/integrator/dataservices/core/DataServiceFault.java
+++ b/components/data/data-services/org.wso2.micro.integrator.dataservices.core/src/main/java/org/wso2/micro/integrator/dataservices/core/DataServiceFault.java
@@ -197,7 +197,7 @@ public class DataServiceFault extends Exception {
 			buff.append("Current Request Name: " + this.getCurrentRequestName() + "\n");
             getPropertyMap().put(DBConstants.FaultParams.CURRENT_REQUEST_NAME, this.getCurrentRequestName());
 		}
-		if (this.getCurrentParams() != null && !("true".equalsIgnoreCase(DBUtils.getCurrentParamsDisabled()))) {
+		if (this.getCurrentParams() != null && !("true".equalsIgnoreCase(DBUtils.getCurrentParamsDisabledProperty()))) {
 			buff.append("Current Params: " + this.getCurrentParams() + "\n");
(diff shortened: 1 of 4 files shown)


Nirothipan:
since it says the issue says that server might hang if there are no. of params. shouldn't we enable this by default

components/data/data-services/org.wso2.micro.integrator.dataservices.core/src/main/java/org/wso2/micro/integrator/dataservices/core/DataServiceFault.java
```diff
@@ -197,7 +197,8 @@ public String getFullMessage() {
 			buff.append("Current Request Name: " + this.getCurrentRequestName() + "\n");
             getPropertyMap().put(DBConstants.FaultParams.CURRENT_REQUEST_NAME, this.getCurrentRequestName());
 		}
-		if (this.getCurrentParams() != null) {
+		String isCurrentParamsDisabled = System.getProperty(DBConstants.DISABLE_CURRENT_PARAMS_IN_LOG);
```
arunans23, line 200: Is it possible to read the property at server startup and not for every fault? It might affect the performance. WDYT?

chanikag:
> since the issue says that server might hang if there are no. of params. shouldn't we enable this by default

No, generally, this is a useful information when it comes to find the root cause of the failure. The server hanging was happened when there are large number of records which were failing. As a tuning option we can ask to disable, if someone is processing large number of requests as batch requests.

components/data/data-services/org.wso2.micro.integrator.dataservices.core/src/main/java/org/wso2/micro/integrator/dataservices/core/DataServiceFault.java
chanikag: +1.

Nirothipan:
shall we add this to docs please

chanikag:
> shall we add this to docs please

+1. I have added that in the internal discussion.

chanikag:
> > shall we add this to docs please
> 
> +1. I have added that in the internal discussion.

Added https://github.com/wso2/docs-ei/pull/1693

PR #1750, opened 2020-06-24
Refactor data service resource in management api.

sachithKay:

### DIFF
--- a/components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/DataServiceResource.java
+++ b/components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/DataServiceResource.java
@@ -27,19 +27,26 @@ import org.apache.synapse.core.axis2.Axis2MessageContext;
 import org.json.JSONObject;
 import org.wso2.micro.integrator.dataservices.common.DBConstants;
 import org.wso2.micro.integrator.dataservices.core.DBUtils;
+import org.wso2.micro.integrator.dataservices.core.description.config.Config;
+import org.wso2.micro.integrator.dataservices.core.description.operation.Operation;
 import org.wso2.micro.integrator.dataservices.core.description.query.Query;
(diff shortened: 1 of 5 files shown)


components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/DataServiceResource.java
```diff
@@ -163,4 +171,33 @@ private ServiceMetaData getServiceMetaData(DataService dataService) throws Excep
...
+    private List<ResourceInfo> getResources(DataService dataService) {
+
+        Set<Resource.ResourceID> resourceIDS = dataService.getResourceIds();
+        List<ResourceInfo> resourceList = new ArrayList<>();
+        for (Resource.ResourceID id : resourceIDS) {
+            ResourceInfo resourceInfo = new ResourceInfo();
+            Resource resource = dataService.getResource(id);
+            resourceInfo.setResourcePath(id.getPath());
+            resourceInfo.setResourceMethod(id.getMethod());
+            resourceInfo.setResourceQuery(resource.getCallQuery().getQuery().getQueryId());
```
chanikag, line 184: Do we need query info in the dashboard view? It is kind of an internal information. Isn't it?

sachithKay: We need to show which query is related to the resource right? Please refer below.
Screen Shot 2020-06-25 at 9 46 47 AM

chanikag: It is an internal detail. We do not expose that from WSDL, etc. Instead we should expose query params.

sachithKay: Modified as per the discussion. Added query params as well

Issue #1751, opened 2020-06-25
[External user store][Ldap][Management API]Get Users return 404


isuruuy429:
**Description:**
When it is connected to an external user store (ldap), get users return 404. 
Doc: https://ei.docs.wso2.com/en/7.1.0/micro-integrator/administer-and-observe/working-with-management-api/#get-users

**Steps to reproduce:**
1. Create the following ldap user store. 
[userstore-ldif.zip](https://github.com/wso2/micro-integrator/files/4829475/userstore-ldif.zip)

2. Add the below configurations to the deployment.toml file. 
```
[internal_apis.file_user_store]
enable = false

[user_store]
type = "read_only_ldap"
class = "org.wso2.micro.integrator.security.user.core.ldap.ReadOnlyLDAPUserStoreManager"
connection_url = "ldap://localhost:10389"
connection_name = "uid=admin,ou=system"
connection_password = "secret"
user_search_base = "ou=system"
```

3. Start the micro integrator as below. 
`./micro-integrator.sh -DenableManagementApi`

4. Login to the server as below by obtaining the access token. 
```
curl -X GET "https://localhost:9164/management/login" -H "accept: application/json" -H "Authorization: Basic YWRtaW46c2VjcmV0" -k -i
```

5. Now try to invoke the users and add users as below. 
```
curl -X GET "https://localhost:9164/management/users?pattern=”*us*”&role=”role”" -H "accept: application/json" -H "Authorization: Bearer %AccessToken%" -k -i
```

```
curl -X POST -d @user "https://localhost:9164/management/users" -H "accept: application/json" -H "Content-Type: application/json" -H "Authorization: Bearer %AccessToken% " -k -i
```
https://ei.docs.wso2.com/en/7.1.0/micro-integrator/administer-and-observe/working-with-management-api/#add-users

**Expected**: List the users or add users
**Actual**: Getting the following error. 

```
HTTP/1.1 404 Not Found
[...]

PR #1752, opened 2020-06-25
Fixes issues in users stores

sdkottegoda:
## Purpose
This PR contains 2 fixes:
1. Removes searching for hybrid user roles upon retrieving the role list of user since those are not added. Resolves: https://github.com/wso2/micro-integrator/issues/1751
2. Avoid printing error logs for known method calls.

### DIFF
--- a/components/org.wso2.micro.integrator.security/src/main/java/org/wso2/micro/integrator/security/user/core/system/SystemUserRoleManager.java
+++ b/components/org.wso2.micro.integrator.security/src/main/java/org/wso2/micro/integrator/security/user/core/system/SystemUserRoleManager.java
@@ -64,7 +64,7 @@ public class SystemUserRoleManager {
 
     public void addSystemRole(String roleName, String[] userList) throws UserStoreException {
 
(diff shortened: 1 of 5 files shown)


components/org.wso2.micro.integrator.security/src/main/java/org/wso2/micro/integrator/security/user/core/system/SystemUserRoleManager.java
```diff
@@ -64,7 +64,7 @@ public SystemUserRoleManager(DataSource dataSource, int tenantId) throws UserSto
 
     public void addSystemRole(String roleName, String[] userList) throws UserStoreException {
 
-        log.error("This Functionality is not available in WSO2 Micro Integrator");
+        log.debug("This Functionality is not available in WSO2 Micro Integrator", new Throwable());
```
chanikag, line 67: Can't we use the same Util method here?

sdkottegoda: done. Thanks :)

PR #1765, opened 2020-06-29
Change default to false for JDBC Userstore ReadOnly property and fix typos in User store error messages

arunans23:
- Change default to false for JDBC Userstore ReadOnly property. This will help creating default admin user when setting up a JDBC user store.
- Fix typos in User store error messages


Nirothipan:
> ## Purpose
> > Describe the problems, issues, or needs driving this feature/fix and include links to related issues in the following format: Resolves issue1, issue2, etc.
> 
> * Change default to false for JDBC Userstore ReadOnly property. This will help creating default admin user when setting up a JDBC user store.

I think <AddAdmin> prop should do it in user-mgt.xml .. Isn't that the case ?

arunans23:
Doc PR : https://github.com/wso2/docs-ei/pull/1704

PR #1767, opened 2020-06-29
Fix MI-536:Invalidate registry cache upon update on properties file

chanikag:
## Purpose
This is to fix https://github.com/wso2/micro-integrator/issues/536

MicroIntegratorRegistryListener is registered to monitor file changes in the registry.

### DIFF
--- a/components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
+++ b/components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
@@ -48,8 +48,12 @@ import java.net.URL;
 import java.net.URLConnection;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import javax.activation.DataHandler;
 import javax.xml.stream.XMLInputFactory;
 import javax.xml.stream.XMLStreamException;
@@ -96,6 +100,8 @@ public class MicroIntegratorRegistry extends AbstractRegistry {
      */
     private int registryProtocol = FILE;
 
+    private static Map<String, Long> resourceLastModifiedMap = new HashMap<String, Long>();
+
 
     public MicroIntegratorRegistry() {
         //default registry is file system based resided in carbon home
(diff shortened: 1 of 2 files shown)


components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistryListener.java
```diff
@@ -0,0 +1,159 @@
...
+        });
+    }
+
+    public void run() {
+        try {
+            WatchKey key;
+            while ((key = watcher.take()) != null) {
+                Path dir = keys.get(key);
+                if (dir == null) {
+                    log.error("WatchKey for listening to registry resources changes not recognized!!");
```
sdkottegoda, line 92: Do we need to do something like invalidating the key or something? Basically will the listener recover if it comes to this state?

components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
```diff
@@ -115,6 +121,34 @@ public MicroIntegratorRegistry() {
...
         initSecurityRepo();
+        initRegistryListener(defaultFSRegRoot);
+    }
+
+    private void initRegistryListener(String regRoot) {
+        try {
+            MicroIntegratorRegistryListener watcher = new MicroIntegratorRegistryListener(Paths.get(regRoot));
+            ExecutorService executor = Executors.newSingleThreadExecutor();
+            executor.submit(watcher);
+            executor.shutdown();
```
sdkottegoda, line 132: Is it correct to shutdown the executor here?

components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
```diff
@@ -115,6 +121,34 @@ public MicroIntegratorRegistry() {
...
+        } catch (IOException e) {
+            log.error(e);
```
sdkottegoda, line 134: Don't we need to throw an exception and interrupt the server startup?

components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistryListener.java
chanikag: Yes, the task is continued after that.. So the listener will take another key when an event happen.

components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
chanikag: Yes, this is to indicate that there is no task anymore to submit to the executor

components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
chanikag: +1. Fixed

components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
```diff
@@ -367,7 +403,7 @@ public RegistryEntry getRegistryEntry(String key) {
             entryEmbedded.setType(MicroIntegratorRegistryConstants.FILE);
             entryEmbedded.setDescription("Resource at : " + url.toString());
             entryEmbedded.setLastModified(urlc.getLastModified());
-            entryEmbedded.setVersion(urlc.getLastModified());
+            entryEmbedded.setVersion(getLastModifiedTimestamp(resolveRegistryURI(key), urlc));
```
Vathsan, line 406: Would only setting the version is enough? Do not we need to remove the particular key from [synapseConfiguration](https://github.com/wso2/wso2-synapse/blob/e01af546918c60e846fea1898fcc91431b19e90f/modules/core/src/main/java/org/apache/synapse/config/SynapseConfiguration.java#L929)?

chanikag: Yes, it checks for the version https://github.com/wso2/wso2-synapse/blob/master/modules/core/src/main/java/org/apache/synapse/registry/AbstractRegistry.java#L96

Issue #1800, opened 2020-07-07
Provide support to write to ldap user store


Nirothipan:
$subject. 
We need to provide support to write to LDAP user store , now we can read only.

sdkottegoda:
ReadWriteLDAPUserStoreManager is already present in the code. We need to provide the capability to configure it and test it with an LDAP.

PR #1801, opened 2020-07-07
Provide support to enable read write ldap in toml

Nirothipan:
Fixes https://github.com/wso2/micro-integrator/issues/1800


distribution/src/resources/config-tool/deployment-full.toml
```diff
@@ -47,7 +47,7 @@ properties.isCascadeDeleteEnabled = true    #inferred
 
 [user_store]
 #usr-mgt.xml
-type = "read_only_ldap" # inferred default read_only_ldap
+type = "read_only_ldap" # inferred default read_only_ldap # OR  read_write_ldap
 class = "org.wso2.micro.integrator.security.user.core.ldap.ReadOnlyLDAPUserStoreManager" # inferred
```
sdkottegoda, line 51: let's remove this entry, may be in a separate commit. This is an advanced configuration.

Nirothipan: since this is full toml , does it matters ?

Issue #1796, opened 2020-07-06
Dataservices sample testcases are disabled.


chanikag:
**Description:**
As the correct package name is not added in [1], the sample testcases hasn't run.

[1] https://github.com/wso2/micro-integrator/blob/master/integration/dataservice-hosting-tests/tests-integration/tests/src/test/resources/testng.xml#L177

PR #1828, opened 2020-07-10
Enable dataservice sample testcases

chanikag:
## Purpose
This is to fix https://github.com/wso2/micro-integrator/issues/1796

The scenarios which are covered under services are deleted


chanikag:
https://github.com/wso2/micro-integrator/pull/1849

PR #1864, opened 2020-07-16
Add root resource for management api and skip authentication for it

Nirothipan:
$subject

### DIFF
--- a/components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RootResource.java
+++ b/components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RootResource.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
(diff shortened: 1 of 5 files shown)


components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/RootResource.java
```diff
@@ -0,0 +1,55 @@
...
+        return methods;
+    }
+
+    @Override
+    public boolean invoke(MessageContext msgCtx, org.apache.axis2.context.MessageContext axisMsgCtx,
+                          SynapseConfiguration synapseConfiguration) {
+
+        CarbonServerConfigurationService serverConfig = CarbonServerConfigurationService.getInstance();
+        String miVersion = serverConfig.getServerVersion();
+        String response = "{ \n WSO2 Micro Integrator :" + miVersion + " \n ManagementApi : Active\n}";
```
sdkottegoda, line 51: Don't we need the ','(comma) before 'ManagementApi'

Nirothipan: This will be displayed as follows and  don't think we need to show with comma
Screenshot from 2020-07-16 13-42-33

Issue #1763, opened 2020-06-29
Capability to enable tracing dynamically


sdkottegoda:
**Description:**
This issue is for dynamically enabling tracing for synapse artifacts. There are 2 options we need to think about.

1. Enabling tracing globally.
2. Enabling tracing per artifact.

IMO, enabling tracing globally will be the better approach since it'll be a one time task. However, the performance impact in this approach will be higher. Allowing bother options will be the best if time permits.

tishan89:
We will be going ahead with option two. Also please consider that there shouldn't be a requirement to re-deploy the artifact

Nirothipan:
Management API support is added in 
https://github.com/wso2/micro-integrator/pull/1786
https://github.com/wso2/micro-integrator/pull/1785
https://github.com/wso2/micro-integrator/pull/1779
https://github.com/wso2/micro-integrator/pull/1782
https://github.com/wso2/micro-integrator/pull/1784

Nirothipan:
Dashboard support is added with https://github.com/wso2/product-mi-tooling/pull/52

PR #1869, opened 2020-07-16
Add dynamically enabling tracing support for endpoints

Vathsan:
## Purpose
This PR adds support to dynamically enable/disable tracing for endpoints

Related issue https://github.com/wso2/micro-integrator/issues/1763

### DIFF
--- a/components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/EndpointResource.java
+++ b/components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/EndpointResource.java
@@ -43,8 +43,8 @@ import java.util.Set;
 
 import static org.wso2.micro.integrator.management.apis.Constants.ACTIVE_STATUS;
 import static org.wso2.micro.integrator.management.apis.Constants.INACTIVE_STATUS;
-import static org.wso2.micro.integrator.management.apis.Constants.NAME;
 import static org.wso2.micro.integrator.management.apis.Constants.STATUS;
+import static org.wso2.micro.integrator.management.apis.Constants.TRACING;
 
 public class EndpointResource implements MiApiResource {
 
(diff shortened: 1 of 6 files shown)


components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/EndpointResource.java
```diff
@@ -87,7 +88,7 @@ public boolean invoke(MessageContext messageContext,
                 if (payload.has(NAME) && payload.has(STATUS)) {
                     changeEndpointStatus(axis2MessageContext, synapseConfiguration, payload);
                 } else {
-                    Utils.setJsonPayLoad(axis2MessageContext, Utils.createJsonErrorObject("Missing parameters in payload"));
+                    handleTracing(payload.get(NAME).getAsString(), messageContext, axis2MessageContext);
```
sdkottegoda, line 91: Don't we need to check if the tracing element is present in the payload?

sdkottegoda: What kind of a payload are we expecting here?

Vathsan: We handle both updating trace, an erroneous scenario inside handleTracing() method. We check for the tracing element in the common handleTracing() method we have in the Utils.java class. https://github.com/wso2/micro-integrator/blob/master/components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/Utils.java#L124.

We use the same payload structure as of the one we use to activate/deactivate endpoints.

`curl -X POST https://localhost:9164/management/endpoints  -H 'authorization: Bearer TOKEN' -H 'content-type: application/json' -d '{"name": "endpointName", "trace": "enable"} -k`

PR #1874, opened 2020-07-17
Parameterize variables for tests

Nirothipan:
Modify the framework so that tests can run irrespective of the environment.

If need to change db, set the following system properties ( these default mysql if not set )

- db_url
- db_user
- db_pwd
- db_driver

If the tables are created before hand, set below system property to True

- dbProvided

### DIFF
--- a/integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/db/manager/DatabaseManager.java
+++ b/integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/db/manager/DatabaseManager.java
@@ -20,6 +20,7 @@ package org.wso2.esb.integration.common.extensions.db.manager;
 
 import com.moandjiezana.toml.Toml;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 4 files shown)


integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/db/manager/DatabaseManager.java
```diff
@@ -281,12 +282,27 @@ private void parseToml(String filePath) throws AutomationFrameworkException {
...
         } catch (Exception ex) {
             throw new AutomationFrameworkException(ex);
         }
     }
 
+    private String resolveValue(String tomlValue, String defValue) {
+
+        if (tomlValue.startsWith("$sys{")) {
+            String value = defValue;
+            String envVariableName = tomlValue.substring(5, tomlValue.length() - 1);
```
sachithKay, line 294: I think we can easily use StringUtils. substringBetween(tomlValue, "$sys{", "}")

Nirothipan: done I've easily used it : )

Issue #1903, opened 2020-07-21
Data mapper fails in MI when DMC file contains single line comments


dulanjalidilmi:
**Description:**
When the dmc file contains a single line comment(//) the Datamapper fails with a similar stack trace as follows.

```
[2020-07-09 10:28:07,274] ERROR {org.wso2.carbon.mediator.datamapper.DataMapperMediator} - DataMapper mediator : mapping failed Error while reading input stream. Script engine unable to execute the script javax.script.ScriptException: <eval>:2:0 Expected } but found eof


^ in <eval> at line number 2 at column number 0
	at org.wso2.carbon.mediator.datamapper.engine.input.readers.JSONInputReader.read(JSONInputReader.java:62)
	at org.wso2.carbon.mediator.datamapper.engine.input.InputBuilder.buildInputModel(InputBuilder.java:59)
	at org.wso2.carbon.mediator.datamapper.engine.core.mapper.MappingHandler.doMap(MappingHandler.java:90)
	at org.wso2.carbon.mediator.datamapper.DataMapperMediator.transform(DataMapperMediator.java:390)
	at org.wso2.carbon.mediator.datamapper.DataMapperMediator.mediate(DataMapperMediator.java:301)
	at org.apache.synapse.mediators.AbstractListMediator.mediate(AbstractListMediator.java:109)
	at org.apache.synapse.mediators.AbstractListMediator.mediate(AbstractListMediator.java:71)
	at org.apache.synapse.mediators.base.SequenceMediator.mediate(SequenceMediator.java:158)
	at org.apache.synapse.mediators.base.SequenceMediator.mediate(SequenceMediator.java:214)
	at org.apache.synapse.mediators.AbstractListMediator.mediate(AbstractListMediator.java:109)
	at org.apache.synapse.mediators.base.SequenceMediator.mediate(SequenceMediator.java:281)
	at org.apache.synapse.core.axis2.Axis2SynapseEnvironment.mediateFromContinuationStateStack(Axis2SynapseEnvironment.java:807)
	at org.apache.synapse.core.axis2.Axis2SynapseEnvironment.injectMessage(Axis2SynapseEnvironment.java:305)
	at org.apache.synapse.core.axis2.SynapseCallbackReceiver.handleMessage(SynapseCallbackReceiver.java:578)
	at org.apache.synapse.core.axis2.SynapseCallbackReceiver.receive(SynapseCallbackReceiver.java:195)
[...]

PR #1908, opened 2020-07-22
Fix Data Mapper failure in MI when DMC file contains single line comments

dulanjalidilmi:
Fix wso2/micro-integrator#1903
When there is a single line comment MI identifies the rest of the content after the comment section as well as a comment. Therefore appended a new line after each line reads.


components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
```diff
@@ -1005,8 +1005,10 @@ private OMNode readNonXML (URL url) throws IOException {
                 StringBuilder strBuilder = new StringBuilder();
                 try (BufferedReader bReader = new BufferedReader(new InputStreamReader(inputStream))) {
                     String line;
+                    String newLineChar = System.getProperty("line.separator");
```
Nirothipan, line 1008: shall we make this as a class level static variable , we might not have to read each time when we execute this method

dulanjalidilmi: fixed.

PR #1915, opened 2020-07-23
Add default http connection timeout

Nirothipan:
$subject


sdkottegoda:
should we add this to the deployment.toml as well as a commented out entry?

Nirothipan:
> should we add this to the deployment.toml as well as a commented out entry?

Added

PR #1925, opened 2020-07-27
Make management port configurable

niruhan:
## Purpose
$subject

### DIFF
--- a/integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
+++ b/integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
@@ -144,7 +144,12 @@ public class CarbonServerManager {
                 }
             }));
 
-            managementPort = 9154 + portOffset;
+            if (commandMap.get("managementPort") != null) {
+                managementPort = Integer.parseInt(commandMap.get("managementPort")) + portOffset;
+            } else {
+                managementPort = 9763 + portOffset;
+            }
+
             waitTill(() -> !isRemotePortInUse("localhost", managementPort), 180, "startup");
 
             if (!isRemotePortInUse("localhost", managementPort)) {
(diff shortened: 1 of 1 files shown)


integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
```diff
@@ -144,7 +144,12 @@ public synchronized void startServerUsingCarbonHome(String carbonHome, Map<Strin
                 }
             }));
 
-            managementPort = 9154 + portOffset;
+            if (commandMap.get("managementPort") != null) {
+                managementPort = Integer.parseInt(commandMap.get("managementPort")) + portOffset;
+            } else {
+                managementPort = 9763 + portOffset;
```
Nirothipan, line 150: shall we make MI port as default ( 9154 ) .. then we don't need to edit MI , automation xml files

integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
```diff
@@ -144,7 +149,12 @@ public synchronized void startServerUsingCarbonHome(String carbonHome, Map<Strin
                 }
             }));
 
-            managementPort = 9154 + portOffset;
+            if (StringUtils.isNotEmpty(commandMap.get("managementPort"))) {
```
Nirothipan, line 152: better to retrieve both from system prop . then no need to edit automation xml to run tests with EI ..

niruhan: From where we will set the System.property in connectors? (Classes inside connector and integration base gets hit after server starts)

Nirothipan: then how is that handled in scriptName = System.getProperty("startupScript");

niruhan: it's resolving to null and reading the value from scriptName = commandMap.get("startupScript");. I can remove the scriptName = System.getProperty("startupScript") part.

PR #1932, opened 2020-07-30
Read script name and mgt port from system prop

niruhan:
## Purpose
$subject


integration/automation-extensions/src/main/java/org/wso2/esb/integration/common/extensions/carbonserver/CarbonServerManager.java
```diff
@@ -91,7 +91,7 @@ public synchronized void startServerUsingCarbonHome(String carbonHome, Map<Strin
             File commandDir = new File(carbonHome);
 
             log.info("Starting server ... ");
-            scriptName = commandMap.get("startupScript");
+            scriptName = System.getProperty("startupScript");
```
Nirothipan, line 94: if this is null we'll have to take from command map . else MI tests would fail

PR #1938, opened 2020-08-01
Reformat deployment toml jms configurations

madushadhanushka:
Refactor the deployment toml file jms configuration misplacement.
Fix truststore primary key issue


Nirothipan:
Could you please check and update other configs as well , eg:- rabbitmq and etc

PR #1976, opened 2020-09-29
Throw exception when target property name is empty in DSS Call Mediator

Vathsan:
## Purpose
In DSS Call Mediator, when the target type is set to property, we need to configure the property name. With dynamic property name support, the name of the configured target property may result in null. In that case, we print a warning log and continue the mediation.

From this PR we throw an exception if the configured dynamic property name becomes null.


components/mediation/mediators/dataservices-mediator/org.wso2.micro.integrator.mediator.dataservice/src/main/java/org/wso2/micro/integrator/mediator/dataservice/DataServiceCallMediator.java
```diff
@@ -222,7 +222,8 @@ private void dispatchToService(org.apache.axis2.context.MessageContext axis2Mess
                 if (targetDynamicName != null) {
                     targetPropertyName = targetDynamicName.evaluateValue(messageContext);
                     if (StringUtils.isEmpty(targetPropertyName)) {
-                        log.warn("Evaluated value for " + targetPropertyName + " is empty");
+                        handleException("Target type is set to property. Evaluated value for target " +
+                                "property name is null/empty", messageContext);
```
chanikag, line 226: Shall we change this to null or empty?

Vathsan: Fixed

Nirothipan (approved):

Issue #2017, opened 2020-11-25
Management APIs for deploy/un-deploy Carbon Applications


sajithaliyanage:
$subject. 

According to the VM based CICD pipeline of the Micro Integrator in the APIM4 release, we have to implement the following admin services to deploy and undeploy Carbon application(s) inside the server. 

- **/management/capp-deployer**
Attached Carbon Application(s) in the request will be deployed into the `carbonapps` directory.

- **/management/capp-undeployer**
cAppNamePattern Carbon Application(s) will be deleted from the `carbonapps` directory.
```json
{
    "cAppNamePattern" : "CAPP_NAME_PATTERN"
}
```

Relates to https://github.com/wso2/micro-integrator/issues/1241

arunans23:
We have decided not to implement this feature. Changing the Integration logic while the server is running should be minimized as much as possible.

VirajHasitha:
The documentation [1] and [2] still not updated with the configuration information. 

### Instructions to use maven-car-deploy-plugin in MI.

1. Navigate to composite exporter project pom.xml file in the maven multi module project created in the integration studi.

2. Find the maven-car-deploy-plugin in the plugins section. 

3. Then add the following plugin configurations

eg:-
```
<plugin>
	<groupId>org.wso2.maven</groupId>
	<artifactId>maven-car-deploy-plugin</artifactId>
	<version>5.2.36</version>
	<extensions>true</extensions>
	<configuration>
		<carbonServers>
			<CarbonServer>
				<trustStorePath>path_to_wso2carbon.jks</trustStorePath>
				<trustStorePassword>wso2carbon</trustStorePassword>
				<trustStoreType>JKS</trustStoreType>
				<serverUrl>https://localhost:9164</serverUrl>
				<userName>admin</userName>
				<password>admin</password>
				<operation>deploy</operation>
			</CarbonServer>
		</carbonServers>
	</configuration>
</plugin>
```

Note: operation can be deploy or undeploy 

4. Then run **mvn clean deploy -Dmaven.deploy.skip=true -Dmaven.car.deploy.skip=false** which will add the corresponding capp to the remote server's deploy folder which will hot deploy.

[1]. https://ei.docs.wso2.com/en/latest/micro-integrator/administer-and-observe/working-with-management-api/#add-carbon-application

[2]. https://apim.docs.wso2.com/en/latest/observe/mi-observe/working-with-management-api/

Thanks,
Viraj

PR #2018, opened 2020-11-25
Add CApp deploy/undeploy services

sajithaliyanage:
$subject

Fixes https://github.com/wso2/micro-integrator/issues/2017


arunans23:
We have decided not to implement this feature. Changing the Integration logic while the server is running should be minimized as much as possible.

Issue #2397, opened 2021-08-26
Expose proxy services via service Catalog


sdkottegoda:
**Description:**
Currently, only Rest APIs are shown Service Catalog. This task is to expose proxy services to the Service Catalog as Passthrough SOAP endpoint.

The suggested approach for implementing this feature includes the following subtasks.

- [x] https://github.com/wso2/integration-studio/issues/882
- [x] #2414
- [x] https://github.com/wso2/product-apim/issues/11656
- [x] https://github.com/wso2/product-apim/issues/11746
- [x] https://github.com/wso2/docs-apim/issues/4727
- [ ] https://github.com/wso2/product-apim/issues/11744

**Affected Product Version:**
APIM 4.1.0
MI 4.1.0
Integration Studio 8.1.0

sajithaliyanage:
Issue: https://github.com/wso2/integration-studio/issues/888 needs to be addressed along with this feature for migrating users.

nandika:
Only UI tests is pending as UI test framework is under development.

PR #2440, opened 2021-09-24
Adding proxy service support for Service Catalog

sajithaliyanage:
- Iterates through proxy service metadata YAML
- Creates definition.wsdl file for each deployed proxy service using the given metadata YAML
- Sets server.hostname and http/https transport listner ports default when {MI_HOST} and {MI_PORTS} placeholder values are not defined 
- Publishes definition.wsdl and metadata.yaml file included archive to the APIM runtime

Related Issue: https://github.com/wso2/micro-integrator/issues/2414
Related EPIC: https://github.com/wso2/micro-integrator/issues/2397


components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
```diff
@@ -85,17 +87,31 @@
...
-            String host = resolver.resolve();
-            resolver.setVariable(MI_PORT);
-            String port = resolver.resolve();
-            currentUrl = currentUrl.replace(HOST, host).replace(PORT, port);
+        if (currentUrl.contains(HOST) || currentUrl.contains(PORT)) {
+            Map<String, Object> configParser = ConfigParser.getParsedConfigs();
+            String envHost = System.getenv(MI_HOST);
+            // update the {MI_HOST} with server.hostname value
+            if (currentUrl.contains(HOST) && (envHost == null || envHost.isEmpty())) {
+                envHost = (String) configParser.get(SERVER_HOSTNAME);
```
sdkottegoda, line 98: Shall we get the hostname as in https://github.com/wso2/micro-integrator/blob/57715992edaacfd50a40036baa2b4f467307e218/components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/internal/CoreServerInitializer.java#L488 ?
If it makes more sense, we can move this to a common method in MicroIntegratorBaseUtils

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
```diff
@@ -85,17 +87,31 @@
...
+            }
+            String envPort = System.getenv(MI_PORT);
```
sdkottegoda, line 100: Let's not resolve these per artifact. Let's keep 4 pre-resolved hosts for http and https (for both usages of host+port and url) and used them here based on whether the url is parameterized.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
```diff
@@ -85,17 +87,31 @@
...
+            // update the {MI_PORT} with server listener ports + offset values
+            if (currentUrl.contains(PORT) && (envPort == null || envPort.isEmpty())) {
+                int portOffset = Integer.parseInt(System.getProperty(SERVER_PORT_OFFSET));
+                if (currentUrl.startsWith("https")) {
+                    int httpsPort = Integer.parseInt((String) configParser.get(HTTPS_LISTENER_PORT));
```
sdkottegoda, line 105: Let's use the existing ways of reading the value from the axis2.xml

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
```diff
@@ -85,17 +87,31 @@
...
+                    envPort = String.valueOf(httpsPort + portOffset);
+                } else {
+                    int httpPort = Integer.parseInt((String) configParser.get(HTTP_LISTENER_PORT));
+                    envPort = String.valueOf(httpPort + portOffset);
+                }
+            }
+            currentUrl = currentUrl.replace(HOST, envHost).replace(PORT, envPort);
         } else if (currentUrl.contains(URL)) {
+            SystemResolver resolver = new SystemResolver();
```
sdkottegoda, line 114: Why have we used the SystemResolver for this and pure Syste.getEnv for the previous case? Shall we be consistent?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
```diff
@@ -372,6 +388,11 @@ private static boolean processMetadata(File tempDir, File metadataFolder, File m
                                            Map<String, String> md5MapOfAllService) throws IOException
             , ResolverException, NoSuchAlgorithmException {
         String metaFileName = metadataYamlFolder.getName();
+        // check metadata is relevant to proxy-service
+        if (metaFileName.contains(PROXY_SERVICE_SUFFIX) && !(new File(metadataFolder,
+                metaFileName.replaceAll(METADATA_FOLDER_STRING, SWAGGER_FOLDER_STRING))).exists()) {
+            return processProxyServiceMetadata(tempDir, metadataYamlFolder, md5MapOfAllService);
+        }
         String APIName = metaFileName.substring(0, metaFileName.indexOf(METADATA_FOLDER_STRING));
```
sdkottegoda, line 396: What if the metadataFileName contains the PROXY_SERVICE_SUFFIX and the swagger folder exists? Are we going to treat it as an API? IMO, this is wrong, we should not allow creation of such a proxy in the integration studio

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
```diff
@@ -372,6 +388,11 @@ private static boolean processMetadata(File tempDir, File metadataFolder, File m
                                            Map<String, String> md5MapOfAllService) throws IOException
             , ResolverException, NoSuchAlgorithmException {
         String metaFileName = metadataYamlFolder.getName();
+        // check metadata is relevant to proxy-service
+        if (metaFileName.contains(PROXY_SERVICE_SUFFIX) && !(new File(metadataFolder,
```
sdkottegoda, line 392: would endsWith be better than contains?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
```diff
@@ -655,4 +754,64 @@ private static String getFileChecksum(File file) throws NoSuchAlgorithmException
...
+     *
+     * @param yamlFile location of the yamlFile.
+     * @return URL of the metadata service.
+     */
+    private static String getProxyServiceUrlFromMetadata(File yamlFile) {
+        Yaml yaml = new Yaml();
+        String currentServiceUrl = null;
+        try (InputStream yamlStream = new FileInputStream(yamlFile)) {
+            Map<String, Object> obj = (Map<String, Object>) yaml.load(yamlStream);
+            currentServiceUrl = (String) obj.get(SERVICE_URL);
```
sdkottegoda, line 811: We only support 1 service URL, is it? There's no way for us to expose 2 for http and https?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
```diff
@@ -524,12 +617,15 @@ public static boolean checkPreConditions() {
...
         Collection APITable =
                 SynapseConfigUtils.getSynapseConfiguration(
                         org.wso2.micro.core.Constants.SUPER_TENANT_DOMAIN_NAME).getAPIs();
-        if (APITable.isEmpty()) {
-            log.info("Cannot find APIs - aborting the service-catalog uploader");
+        Collection proxyTable =
+                SynapseConfigUtils.getSynapseConfiguration(
+                        org.wso2.micro.core.Constants.SUPER_TENANT_DOMAIN_NAME).getProxyServices();
+        if (APITable.isEmpty() && proxyTable.isEmpty()) {
+            log.info("Cannot find APIs or Proxy Services - aborting the service-catalog uploader");
```
sdkottegoda, line 628: This could probably be a debug log since this is not something significant.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
```diff
@@ -424,6 +445,78 @@ private static boolean processMetadata(File tempDir, File metadataFolder, File m
...
+            // Copy metadata yaml file to the temp location.
+            FileUtils.copyFile(metadataYamlFile, newYamlFile);
+            // Copy wsdl file to the temp location.
+            FileUtils.copyFile(wsdlFile, new File(newMetaFile, wsdlFile.getName()));
+
+            // Add to map to be included in filter formdata field.
+            Map<String, String> md5Map = new HashMap<>();
+            md5Map.put(METADATA_KEY, key);
+            // if API is changed add the previous MD5 sum, if new API add the new MD5 sum.
+            md5Map.put(MD5, StringUtils.isEmpty(md5FromServer) ? newMD5String : md5FromServer);
```
sdkottegoda, line 509: Do we really need a list of maps here? Can't we simply use an object instead of the map and add it to the list? Or atleast limit the size of the map since we know that it is 2?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sajithaliyanage: In previous cases, it checks the `MI_HOST` and `MI_PORT` null check to fetch the system default values (to avoid breaking default behaviour). But for `MI_URL` we don't need to resolve it by default since it is a manually configured placeholder.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sajithaliyanage: We cannot use 'endsWith' here because metaFileName value is `<PROXY_NAME>_proxy_metadata_<VERSION>`

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sajithaliyanage: yes, the Service catalog is designed to use one service URL from the metadata file.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sajithaliyanage: Btw, this log prints when the developer has configured Service catalogue configurations in the deployment.toml file. In the 4.0.0 release, we had this log for APIs. WDYT?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sajithaliyanage: I used the same logic as in APIs https://github.com/wso2/micro-integrator/blob/master/components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java#L416 to maintain the same behaviour.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sajithaliyanage: Here, it checks the swagger **file** exists inside the metadata folder. Folder structures for the APIs and Proxy services will be as follows.

**If the folder is for API metadata:**
- APIName_metadata_1.0.0
      -  APIName_metadata-1.0.0.yaml
      -  APIName_swagger-1.0.0.yaml

**If the folder is for Proxy metadata:**
- ProxyName_proxy_metadata_1.0.0
      -  ProxyName_proxy_metadata-1.0.0.yaml

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sajithaliyanage: Fixed

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sajithaliyanage: Fixed

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sdkottegoda: @tharindu1st Shouldn't this option be available?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sdkottegoda: This will get printed for each task run until the first API is deployed right? Do you see any useful information?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sajithaliyanage: Added Utils methods to get Hostname, HTTP port and HTTPS port.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sajithaliyanage: No, this is a one-time log. +1 to change this as debug log. Fixed

sajithaliyanage: Fixed

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/internal/MicroIntegratorBaseConstants.java
```diff
@@ -46,6 +46,9 @@
     public static final String COMPONENT_REP0_ENV = "COMPONENTS_REPO";
     public static final String AXIS2_REPO_ENV = "AXIS2_REPO";
     public static final String UPDATE_LEVEL = "UPDATE_LEVEL";
+    public static final String TRANSPORT_PORT = "port";
+    public static final String TRANSPORT_HTTP = "http";
+    public static final String TRANSPORT_HTTPS = "https";
```
sdkottegoda, line 51: Do we need to redefine these? Can't we use the existing constants?

sajithaliyanage: Fixed

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/util/MicroIntegratorBaseUtils.java
```diff
@@ -607,4 +618,44 @@ public static SynapseEnvironment getSynapseEnvironment() {
...
+     * Gets Server http listener port.
+     *
+     * @return http port
+     * @throws ResolverException exception
+     */
+    public static int getServerHTTPListenerPort() throws ResolverException {
+        try {
+            return Integer.parseInt(carbonAxisConfigurator.getAxisConfiguration().getTransportsIn()
+                    .get(org.wso2.micro.core.Constants.HTTP_TRANSPORT)
+                    .getParameter(org.wso2.micro.core.Constants.TRANSPORT_PORT).getValue().toString());
```
sdkottegoda, line 640: Shall we resolve the port offset and return?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sdkottegoda: The fact that I am saying is we don't need to do these per artifact. We have have simply have 4 resolved values, MI_ HOST,  HTTP_MI_URL, HTTP_PORT, HTTPS_PORT and replace them depending on the usage.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sdkottegoda: Assume a day where we have changed the SystemResolver logic to lookup system properties as well. The 2 cases will behave differently. So let's make it consistent.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sdkottegoda: As per the offline chat with Tharindu this will not be required.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/utils/ServiceCatalogUtils.java
sajithaliyanage: Done

components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/integrator/core/util/MicroIntegratorBaseUtils.java
sajithaliyanage: Fixed

PR #2645, opened 2022-03-14
Fix admin API message store consumer issue

madushadhanushka:
Fix https://github.com/wso2/product-ei/issues/5548


SanojPunchihewa:
We may need to remove these from the apictl as well. https://github.com/wso2/product-apim-tooling/blob/master/import-export-cli/mi/impl/getMessageStores.go#L38

sachithKay (approved):

PR #2681, opened 2022-03-31
Fix NPE in DataServiceCall mediator

rosensilva:


arunans23:
Didn't we fix this in https://github.com/wso2/micro-integrator/pull/2702?

rosensilva:
fixed with https://github.com/wso2/micro-integrator/pull/2702

PR #2699, opened 2022-04-23
Running JUnit Tests in Parallel with Maven

shisheng-1:
According to [Maven parallel test](https://www.baeldung.com/maven-junit-parallel-tests), we can run tests in parallel.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.


rosensilva:
Hi @shisheng-1 

We have tried this with parallel builds and our current integration test framework does not support this feature. First, we have to isolate all the integration tests so that we can enable them to run in parallel mode.

Thank you for your contribution, we will add the parallel test execution as soon as we make the integration test framework compatible with maven parallel tests.

PR #2714, opened 2022-06-01
[WIP] Fix SSL issue in Netty transport

thuva9872:
## Purpose
HTTP/2 integration tests fail as the Jenkins java 8 build doesn't support TLSv1.3. It needs java8u272 and a later version to support TLSv1.3.

##Goal
Remove TLSv1.3 from the inbound endpoint configuration.


thuva9872:
We are closing this PR as we are upgrading the JDK8 version of Jenkins.

PR #2739, opened 2022-07-26
netty jars version upgrade to fix micro-integrator startup issue in aarch64

chameerar:
## Purpose
WSO2 Micro Gateway was not properly starting on aarch64 architecture. The root cause was regarding to an issue in netty libraries which was [fixed](https://github.com/netty/netty/pull/12498). This change upgrades the netty libraries used in micro-integrator so that the fix provided by netty is reflected.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -577,6 +577,11 @@
             </dependency>
             <dependency>
                 <groupId>io.netty</groupId>
+                <artifactId>netty-transport-native-unix-common</artifactId>
+                <version>${netty.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.netty</groupId>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -575,6 +575,11 @@
                 <artifactId>netty-transport</artifactId>
                 <version>${netty.version}</version>
             </dependency>
+            <dependency>
```
GDLMadushanka, line 578: is this new jar getting bundled ? please check

chameerar: Added required changes to pom.xml files. Please review.

GDLMadushanka (approved):

chameerar:
Fix for - https://github.com/wso2/api-manager/issues/462

PR #2776, opened 2022-10-29
parameterize transport properties in deployment.toml file

madushadhanushka:
Parameterize properties from deployment.toml file.
Frontport https://github.com/wso2/product-apim/pull/12931


distribution/src/resources/config-tool/templates/conf/axis2/axis2_client.xml.j2
```diff
@@ -0,0 +1,298 @@
+<!--
```
sajinieKavindya, line 1: Shall we update the license headers?

sajinieKavindya (approved):

distribution/src/resources/config-tool/templates/conf/axis2/axis2_client.xml.j2
madushadhanushka: This is template file taken from the original axis2_client.

PR #2813, opened 2022-12-12
Create a separate script to run the extensions in MI

arunans23:
$subject


components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
```diff
@@ -201,8 +200,9 @@ public void init(Properties properties) {
                 addConfigProperty(name, value);
             }
         }
-        log.debug("EI lightweight registry is initialized.");
+        log.debug("MI lightweight registry is initialized.");
 
+        initSecurityRepo();
```
isudana, line 205: Whats the reason for moving this?

components/org.wso2.micro.integrator.server/src/main/java/org/wso2/micro/integrator/server/Extensions.java
```diff
@@ -0,0 +1,46 @@
+package org.wso2.micro.integrator.server;
```
isudana, line 1: License header is missing

components/mediation/registry/org.wso2.micro.integrator.registry/src/main/java/org/wso2/micro/integrator/registry/MicroIntegratorRegistry.java
arunans23: Previously, it was not considering the locations that are configured in the registry.xml. It always tries to create the secure-vault folder in the default _MI_HOME/registry/_ location. It will cause issue in a read-only file system.

isudana (approved):

PR #2826, opened 2023-01-10
Handle invalid heartbeat responses

PasanT9:
Introduced a JsonParseException while converting the heartbeat response to a JSONObject. If the response is in incorrect format or empty this will assign a null value for the jsonResponse. When jsonResponse is null MI will skip processing the current heartbeat response and wait for the next scheduled heartbeat query.

### DIFF
--- a/components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/dashboard/HeartBeatComponent.java
+++ b/components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/dashboard/HeartBeatComponent.java
@@ -19,6 +19,7 @@ package org.wso2.micro.integrator.initializer.dashboard;
 
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
 import com.google.gson.JsonParser;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -111,7 +112,7 @@ public class HeartBeatComponent {
                 httpPost.setEntity(entity);
                 CloseableHttpResponse response = client.execute(httpPost);
                 JsonObject jsonResponse = getJsonResponse(response);
-                if (jsonResponse.get("status").getAsString().equals("success")) {
+                if (jsonResponse != null && jsonResponse.get("status").getAsString().equals("success")) {
                     int deployedArtifactsCount = heartbeatPayload.get(CHANGE_NOTIFICATION).getAsJsonObject()
                                                                     .get(DEPLOYED_ARTIFACTS).getAsJsonArray().size();
                     int undeployedArtifactsCount = heartbeatPayload.get(CHANGE_NOTIFICATION).getAsJsonObject()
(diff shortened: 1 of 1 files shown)


chanikag (approved):

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/dashboard/HeartBeatComponent.java
```diff
@@ -123,7 +124,7 @@ public static void invokeHeartbeatExecutorService() {
                     ArtifactUpdateListener.removeFromUpdatedArtifactQueue(updatedArtifactsCount);
                 }
             } catch (IOException | NoSuchAlgorithmException | KeyStoreException | KeyManagementException e) {
```
lankavitharana, line 126: Hi, can I know why we are not catching all the exceptions at least as a fail safe? just asking because say next time some other unseen exception happens, this exact scenario will repeat, any particular reason for not catching everything and continue the thread?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/dashboard/HeartBeatComponent.java
```diff
@@ -205,7 +206,13 @@ public static boolean isDashboardConfigured() {
 
     public static JsonObject getJsonResponse(CloseableHttpResponse response) {
         String stringResponse = getStringResponse(response);
-        return new JsonParser().parse(stringResponse).getAsJsonObject();
+        JsonObject responseObject = null;
+        try {
+            responseObject = new JsonParser().parse(stringResponse).getAsJsonObject();
+        } catch (JsonParseException e) {
+            log.error("Error occurred while parsing the heartbeat response.", e);
```
lankavitharana, line 213: any possibility of changing this log level to debug as well?

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/dashboard/HeartBeatComponent.java
PasanT9: We did this to avoid unexpected exceptions. Catching _Exceptions_ usually not a good practice.

lankavitharana: I'm not an expert, but I would say, that depends on the scenario, like this one for example, we would need the thread to keep retrying whatever happens. Otherwise, the connection between the container and dashboard may disappear without any trace (nothing is even being logged)

PasanT9: Agreed. This might create a issue in the future. I will create a Git issue to discuss this further.

PasanT9: I opened https://github.com/wso2/api-manager/issues/1212 to fix this issue.

components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/dashboard/HeartBeatComponent.java
PasanT9: It will be changed by https://github.com/wso2/micro-integrator/pull/2827

PR #2851, opened 2023-02-02
Add WebContextRoot deployment.toml config

imsuneth:
## Purpose
Fix https://github.com/wso2/api-manager/issues/1300

A new configuration parameter is added to set WebContextRoot in deplyment.toml

## Usage
In deplyment.toml, it can be defined as follows

```
[server]
web_context_root = ""
```

### DIFF
--- a/distribution/src/resources/config-tool/templates/conf/carbon.xml.j2
+++ b/distribution/src/resources/config-tool/templates/conf/carbon.xml.j2
@@ -89,8 +89,11 @@
     <!--
        Webapp context root of WSO2 Carbon management console.
     -->
+    {% if server.webcontexroot is defined %}
+    <WebContextRoot>{{ server.webcontexroot }}</WebContextRoot>
+    {% else %}
     <WebContextRoot>/</WebContextRoot>
-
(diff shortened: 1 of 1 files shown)


distribution/src/resources/config-tool/templates/conf/carbon.xml.j2
```diff
@@ -89,8 +89,11 @@
     <!--
        Webapp context root of WSO2 Carbon management console.
     -->
+    {% if server.webcontexroot is defined %}
+    <WebContextRoot>{{ server.webcontexroot }}</WebContextRoot>
```
madushadhanushka, line 93: Shall we make the parameter with an underscore. `web_context_root`.  Other places use the same convention.

imsuneth: Sure, changed in https://github.com/wso2/micro-integrator/pull/2851/commits/2a8c568673cdbec1884ec256934bf6ed25cfb458

madushadhanushka (approved):

PR #2887, opened 2023-03-15
Add new version of management API

GDLMadushanka:
Add new version to management API which handle url encoded usernames and domains
Fixes wso2/api-manager/issues/1427


components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/ManagementInternalApiV1_1.java
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright (c) 2018, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
arunans23, line 2: Need to change the license header here.

distribution/src/scripts/micro-integrator.sh
```diff
@@ -330,6 +330,7 @@ do
     -DenableReadinessProbe=true \
     -DenableLivenessProbe=true \
     -DenableManagementApi=true \
+    -DenableManagementApiV1_1=true \
```
arunans23, line 333: Can't we make 1.1 as default and have only one system property to enable management APIs?

GDLMadushanka:
Not adding this change since decoding is not needed

PR #2897, opened 2023-04-11
Add new test case to test health check API

madushadhanushka:
## Purpose
Add test cases for https://github.com/wso2/api-manager/issues/1463


rosensilva (approved):

madushadhanushka:
Following test failed in Github build and seems its not related to the test that I have added.

- RegistryResourcesTestCase.testRegistryGetDirectoryExpandToVerifyChanges:665 expected [true] but found [false]

PR #4274, opened 2025-07-16
Introduce new CORS handler that can be engaged per API

arunans23:
$subject

### DIFF
--- a/components/org.wso2.micro.integrator.security/src/main/java/org/wso2/micro/integrator/security/handler/CORSRequestHandler.java
+++ b/components/org.wso2.micro.integrator.security/src/main/java/org/wso2/micro/integrator/security/handler/CORSRequestHandler.java
@@ -1,0 +1,254 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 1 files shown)


components/org.wso2.micro.integrator.security/src/main/java/org/wso2/micro/integrator/security/handler/CORSRequestHandler.java
```diff
@@ -0,0 +1,406 @@
...
+    public void destroy() {
+
+    }
+
+    public boolean handleRequest(MessageContext messageContext) {
+
+        if (log.isDebugEnabled()) {
+            log.debug("CORSRequestHandler: handleRequest called");
+        }
+        if (SynapseCORSConfiguration.getInstance().isEnabled()) {
```
GDLMadushanka: With this, we cannot achieve a custom configuration for a given API and a default config for other APIs. WDYT ?

arunans23: Fixed with https://github.com/wso2/wso2-synapse/pull/2392

components/org.wso2.micro.integrator.security/src/main/java/org/wso2/micro/integrator/security/handler/CORSRequestHandler.java
```diff
@@ -0,0 +1,402 @@
...
+                            selectedResource = resource;
+                            if (selectedResource.getDispatcherHelper()
+                                    .getString() != null && !selectedResource.getDispatcherHelper().getString()
+                                    .contains("/*")) {
+                                break;
+                            }
+                        }
+                    }
+                    if (selectedResource == null) {
+                        handleResourceNotFound(messageContext, Arrays.asList(allAPIResources));
```
GDLMadushanka, line 119: Will this break the normal stat collection flows? Since we are returning from the handler without going into the API.

arunans23: We are not doing any stats collection in the synapse level as well for these scenarios. https://github.com/wso2/wso2-synapse/blob/master/modules/core/src/main/java/org/apache/synapse/api/API.java#L536

PR #4304, opened 2025-08-07
Enable activate/deactivate functionality for Polling Inbound Endpoints

AmilaSamith:
[1] https://mi.docs.wso2.com/en/latest/develop/customizations/creating-custom-inbound-endpoint/

### DIFF
--- a/components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/common/InboundRequestProcessorImpl.java
+++ b/components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/common/InboundRequestProcessorImpl.java
@@ -165,6 +165,10 @@ public abstract class InboundRequestProcessorImpl implements InboundRequestProce
      * Activates the Inbound Endpoint by activating any associated startup controllers
      * or resuming inbound runner threads if no startup controllers are present.
      *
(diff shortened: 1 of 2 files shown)


components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/common/InboundRequestProcessorImpl.java
```diff
@@ -179,7 +183,7 @@ public boolean activate() {
         log.info("Activating the Inbound Endpoint [" + name + "].");
 
         boolean isSuccessfullyActivated = false;
-        if (!startUpControllersList.isEmpty()) {
+        if (this.coordination && !startUpControllersList.isEmpty()) {
```
sajinieKavindya, line 186: I don’t think there are cases where startUpControllersList would be non-empty when coordination is set to false. Therefore, checking the coordination-true condition seems unnecessary. WDYT?

AmilaSamith: Introduced that check to improve clarity in the code since controller list has elements in the case of coordination enabled. If it seems to be redundant we can remove that check. WDYT?

AmilaSamith:
Please note these changes will be send via PR : https://github.com/wso2-extensions/mi-inbound-cdc/pull/34

PR #4341, opened 2025-09-15
Replacing the InboundRunner threads with InboundTasks when the inbound endpoint is running in coordination-false mode

sajinieKavindya:
## Purpose
Instead of running a dedicated `InboundRunner` thread for the inbound endpoint when coordination is set to `false`, we can submit an `InboundTask` to the Task scheduler. Maintaining two separate execution modes is unnecessary and only adds complexity to the code.  

Therefore, the `InboundRunner` is replaced with `InboundTask` when the inbound endpoint runs in coordination-false mode. This change centralizes graceful shutdown handling for inbound endpoints and results in a cleaner, more consistent execution flow.  

Similarly, `OneTimeTriggerInboundRunner` thread is replaced with `OneTimeTriggerInboundTask`.


components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSPollingConsumer.java
```diff
@@ -445,6 +445,12 @@ private void releaseResources(boolean forcefullyClose) {
 
     public void destroy() {
         synchronized (jmsConnectionFactory) {
+            if (destroyed) {
+                if (logger.isDebugEnabled()) {
+                    logger.debug("PollingConsumer already destroyed for Inbound Endpoint: " + name);
+                }
+                return;
+            }
```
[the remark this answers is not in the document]
sajinieKavindya: The "destroy" variable can only be made true inside the destroy() method

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSProcessor.java
```diff
@@ -146,6 +143,10 @@ public void update() {
      */
     @Override
     public void destroy(boolean removeTask) {
+        // Destroying all the polling consumers when the inbound endpoint is removed.
+        for (JMSPollingConsumer pollingConsumer : pollingConsumers) {
+            pollingConsumer.destroy();
+        }
```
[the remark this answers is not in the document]
sajinieKavindya: Whether the polling consumer was destroyed successfully or not will be logged inside pollingConsumer.destroy();

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/common/InboundRequestProcessorImpl.java
```diff
@@ -69,11 +68,7 @@ protected void start(InboundTask task, String endpointPostfix) {
         if (task instanceof GenericTask)  {
             cronExpression = ((GenericTask) task).getPollingConsumer().getCronExpression();
         }
-        if (coordination || (cronExpression != null && !cronExpression.isEmpty())) {
-            handleTask(task, endpointPostfix);
-        } else {
-            startInboundRunnerThread(task, Constants.SUPER_TENANT_DOMAIN_NAME, false, startInPausedMode);
-        }
+        handleTask(task, endpointPostfix);
```
AmilaSamith, line 71: Don't we need to have null and empty check on cronExpression ?

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/common/InboundRequestProcessorImpl.java
```diff
@@ -218,27 +176,19 @@ public boolean activate() {
                     }
                 }
             }
-        } else if (!inboundRunnersThreadsMap.isEmpty()) {
```
AmilaSamith, line 221: Is it enough at least only one startup controller get activated or all should be activated. WDYT?

If all should be activated IMO the logic should be modified as follows, same as in deactivate method.
```
public boolean activate() {
    log.info("Activating the Inbound Endpoint [" + name + "].");
    boolean isSuccessfullyActivated = true;
    
    if (!startUpControllersList.isEmpty()) {
        for (StartUpController sc : startUpControllersList) {
            if (!sc.activateTask()) {
                if (log.isDebugEnabled()) {
                    log.debug("Failed to activate the consumer: " + sc.getTaskDescription().getName());
                }
                isSuccessfullyActivated = false;
            }
        }
    }
    return isSuccessfullyActivated;
}
```

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/common/InboundRequestProcessorImpl.java
sajinieKavindya: That check is done in the handleTask method itself.

AmilaSamith: Noted. check happens within [getTaskDescription()](https://github.com/sajinieKavindya/micro-integrator/blob/326e3c2d98a3c14a4d9fa7b029c03324ff14f7cc/components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/common/InboundRequestProcessorImpl.java#L115)

components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/common/InboundRequestProcessorImpl.java
sajinieKavindya: It was decided that only one startup controller (the task) gets activated is enough to consider the inbound endpoint is in active state. Otherwise, the inbound endpoint could be marked as deactivated even while a consumer is still processing a message, which, in my opinion, is incorrect.

AmilaSamith (approved):
Since we've implemented the improvement to use InboundTask despite of coordination or not, IMO we don't need user input on coordination decisions - that seems unnecessary in my opinion. However, we should thoroughly review the code before finalizing that approach. For now, let's proceed with the current architecture and we can release that change as an update later if needed.

DedunuKarunarathne (approved):

PR #4429, opened 2025-10-17
Introduced an overloaded method to pause the polling consumers during graceful shutdown

sajinieKavindya:
$subject

### DIFF
--- a/components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/generic/GenericPollingConsumer.java
+++ b/components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/generic/GenericPollingConsumer.java
@@ -81,6 +81,8 @@ public abstract class GenericPollingConsumer {
 
     public abstract void pause();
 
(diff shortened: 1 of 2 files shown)


components/mediation/inbound-endpoints/org.wso2.micro.integrator.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/generic/GenericPollingConsumer.java
```diff
@@ -81,6 +81,8 @@ public GenericPollingConsumer(Properties properties, String name, SynapseEnviron
 
     public abstract void pause();
 
+    public abstract void pause(boolean destroyConsumer);
```
AmilaSamith, line 84: Shall we add method comments to reflect the purpose of having two methods. WDYT?

```
# Implement this method to support deactivation of inbound endpoint
public abstract void pause();

# Implement this method to support graceful shutdown of inbound endpoint
public abstract void pause(boolean destroyConsumer);
```

AmilaSamith, line 84: Is this change only needed in polling based consumers?

PR #4980, opened 2026-07-20
Fix Issues in CApp Deployment and File Operations

sajinieKavindya:
This is to validate the files inside the capp before extracting and validate file paths before performing file operations.

### DIFF
--- a/components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/application/deployer/AppDeployerUtils.java
+++ b/components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/application/deployer/AppDeployerUtils.java
@@ -694,11 +694,11 @@ public final class AppDeployerUtils {
 
         zipFile = new ZipFile(sourcePath);
         entries = zipFile.entries();
+        String canonicalDirPath = new File(destPath).getCanonicalPath();
 
         while (entries.hasMoreElements()) {
             ZipEntry entry = (ZipEntry) entries.nextElement();
             String canonicalEntryPath = new File(destPath + entry.getName()).getCanonicalPath();
-            String canonicalDirPath = new File(destPath).getCanonicalPath();
             if (!canonicalEntryPath.startsWith(canonicalDirPath)) {
                 throw new DeploymentException("Entry is outside of the target dir: " + entry.getName());
             }
--- a/components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/application/deployer/AppDeployerUtils.java
+++ b/components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/application/deployer/AppDeployerUtils.java
@@ -697,6 +697,12 @@ public final class AppDeployerUtils {
 
         while (entries.hasMoreElements()) {
             ZipEntry entry = (ZipEntry) entries.nextElement();
+            String canonicalEntryPath = new File(destPath + entry.getName()).getCanonicalPath();
+            String canonicalDirPath = new File(destPath).getCanonicalPath();
+            if (!canonicalEntryPath.startsWith(canonicalDirPath)) {
+                throw new DeploymentException("Entry is outside of the target dir: " + entry.getName());
+            }
+
             // we don't need to copy the META-INF dir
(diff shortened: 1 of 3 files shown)


components/org.wso2.micro.integrator.core/src/main/java/org/wso2/micro/application/deployer/AppDeployerUtils.java
```diff
@@ -697,6 +697,12 @@ private static void extract(String sourcePath, String destPath) throws IOExcepti
 
         while (entries.hasMoreElements()) {
             ZipEntry entry = (ZipEntry) entries.nextElement();
+            String canonicalEntryPath = new File(destPath + entry.getName()).getCanonicalPath();
+            String canonicalDirPath = new File(destPath).getCanonicalPath();
```
arunans23: We can compute the canonical path once outside the while.

sajinieKavindya: We can get the `canonicalDirPath` outside the while loop. Will fix.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/LogFilesResource.java
```diff
@@ -172,22 +172,44 @@ private void sendFaultResponse(org.apache.axis2.context.MessageContext axis2Mess
...
                 return null;
-            } catch (IOException e) {
-                log.error("Error occurred while reading the file : " + logFile + " in : " + logFilePath.toString(), e);
+            }
+
+            // Canonicalize the intended base directory
+            Path logsBasePath = Paths.get(Utils.getCarbonLogsPath()).toRealPath();
+
+            // Resolve and normalize the full target path
+            Path logFilePath = logsBasePath.resolve(logFile).normalize();
```
chanikag, line 187: Better to check the real path for the logFilePath, to avoid any symlinks associated with it.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/CarbonAppResource.java
```diff
@@ -262,21 +262,45 @@ private void populateFileContent(MessageContext synCtx, String cAppName) {
...
+
+            // Canonicalize the intended base directory
+            Path cAppBasePath = Paths.get(Utils.getCAppPath()).toRealPath();
+
+            // Resolve and normalize the full target path
+            Path cAppPath = cAppBasePath.resolve(cAppName).normalize();
+
+            // Validate the resolved path is still within the cApp directory
+            if (!cAppPath.startsWith(cAppBasePath)) {
+                log.error("Path traversal attempt detected for cApp : " + cAppName);
```
chanikag, line 281: Shouldn't these warn logs?

sajinieKavindya: As we are returning from the resource without a successful outcome, providing an Error log is fine IMO.

components/org.wso2.micro.integrator.extensions/org.wso2.micro.integrator.management.apis/src/main/java/org/wso2/micro/integrator/management/apis/LogFilesResource.java
sajinieKavindya: Since logsBasePath is already a real path (symlinks resolved via toRealPath()), and logFilePath is resolved on top of that real base path. The only way a symlink could sneak in is through the logFile string itself, but a symlink cannot be passed as a string parameter. Hence, calling toRealPath() on logFilePath is redundant here.

PR #4996, opened 2026-08-03
Log heartbeat failures at warn level instead of debug

### LINKED ISSUE #4913: Issue with exception handling product-integrator-mi
### Current Limitation

Hi Team,

We noticed that the current exception handling in the product-integrator-mi [1] which silently swallows all exceptions by catching the Exception class and logging only at the debug level. As a result, failures may go unnoticed in production environments where debug logging is typically disabled.

For example, in HeartBeatComponent.java[2]:

} catch (Exception e) {
    log.debug("Error occurred while processing the heartbeat.", e);
}
It would be helpful to handle exceptions more specifically to help with troubleshooting.

[1] https://github.com/wso2/product-integrator-mi/
[2] https://github.com/wso2/product-integrator-mi/blob/9fdecde09d4b7fa4f22108b441218076799dfbfc/components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/dashboard/HeartBeatComponent.java#L70

Best regards,
Ruvindi

### Suggested Improvement

It would be helpful to consider an improvement, such as retaining log.debug for known transient exception types while escalating unexpected exception types or repeated failures beyond a configurable threshold to log.warn or log.error.

### Version

4.1.0

Skynettheone:
## Summary

Fixes #4913.

`HeartBeatComponent.invokeHeartbeatExecutorService()` catches all exceptions from the heartbeat HTTP call to the dashboard/ICP and logs them at `debug` level only. Since debug logging is typically disabled in production, heartbeat failures go completely unnoticed.

## Change

Changed the log level in that catch block from `debug` to `warn`, so failures are visible in production logs by default without needing debug logging enabled. The exception message and full stack trace are still logged, so no diagnostic detail is lost.

This is a minimal, scoped version of the reporter's suggestion (they also proposed a configurable threshold for repeated failures), keeping the change small and focused on the core visibility problem they raised.

## Test plan

- [x] `mvn compile` on `components/org.wso2.micro.integrator.initializer` succeeds with this change (BUILD SUCCESS).

### DIFF
--- a/components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/dashboard/HeartBeatComponent.java
+++ b/components/org.wso2.micro.integrator.initializer/src/main/java/org/wso2/micro/integrator/initializer/dashboard/HeartBeatComponent.java
@@ -114,7 +114,7 @@ public class HeartBeatComponent {
                     log.debug("Error occurred while sending the heartbeat.");
                 }
             } catch (Exception e) {
-                log.debug("Error occurred while processing the heartbeat.", e);
+                log.warn("Error occurred while processing the heartbeat.", e);
             }
         };
         scheduledExecutorService.scheduleAtFixedRate(runnableTask, 1, interval, TimeUnit.SECONDS);


Skynettheone:
Closing this — the underlying issue (#4913) was already resolved and merged via #5005, which makes the same log-level change. No need to duplicate it here.

PR #5054, opened 2026-09-22
Execute ICP management commands tunneled on the heartbeat

manuranga:
## Purpose

An MI behind a load balancer is reachable only outbound: nothing routes from the ICP to a particular replica's management port, so every management view in the ICP console is blank for such a deployment.

The ICP can now queue the management call and hand it over in the response to a heartbeat this node made. This PR is the MI side — the part that performs the call and reports the outcome.

Pairs with the ICP change (wso2/integration-control-plane, linked below). The ICP's `miTunnelEnabled` is **off by default**, and an MI without this change ignores the `commands` array, so the two can be released in either order.

## What it does

`ICPCommandExecutor` performs the command against **this node's own loopback management listener** rather than invoking the management resources directly. That is what keeps it small: every management resource, its authorization handler and its audit logging apply unchanged, and a management API that grows a resource grows this with it for free. The outcome is posted to `POST /icp/commandResult`.

Deliberate choices, each with a failure behind it:

- Commands run on their own single thread, and a batch is refused while the previous one is still running. A slow management resource must not stop this runtime from heartbeating, and the refusal is the back-pressure that stops the ICP handing over work faster than it can be done. A refused batch is re-offered on a later heartbeat, not lost.
- Outcomes are remembered by command id (bounded, 256), so a redelivery after a result lost in flight replays the stored outcome instead of deleting a user twice.
- A command past its deadline is dropped unexecuted — the caller has already been told the read failed, and a late mutation is worse than none. An unparseable deadline counts as expired.
- The path is confined to `/management/` here as well as at the ICP; neither side relies on the other having checked.
- A body travels back as JSON or as a JSON string according to the response's own `Content-Type`, never a parse attempt: a log file whose first line happens to be a bare number is text.
- Every HTTP call this class and the heartbeat make is bounded. An unbounded socket read on the heartbeat thread showed up as a healthy runtime that the console called offline.

The heartbeat loop reschedules itself instead of running at a fixed rate, so it can honour the faster cadence the ICP asks for while a user is working with a management view. Its rescheduling is in a `finally` block: a self-rescheduling loop that skips its own reschedule is a runtime that never reports again.

## Testing

- `ICPCommandExecutorTest` — 10 unit tests over everything the executor decides on its own: path confinement (403), unsupported verb (405), missing params (400), expired and malformed deadlines, replay of a redelivered command id, unreadable payloads, and the text/JSON rule. Every case stops before the listener is contacted, which is also the point — a refusal must not depend on the server being up.
- Module suite green: `mvn -o -pl components/org.wso2.micro.integrator.initializer clean test` → 100 tests, 0 failures.
- Full `mvn clean install` of the product, then end to end against the ICP: tunnelled reads (82 loggers, 11 log files), a tunnelled write (log level change, settled and audited once), and a 137 KB log file downloaded as plain text with no escaping. Also exercised with two MI replicas behind nginx with the management port unreachable — the topology this exists for — where a level change reached both replicas.

> Reviewer notes: this module's tests are skipped by the incremental-build plugin unless the module changed, so `mvn test` can report success having run nothing — use `clean test`. Pass `-Djacoco.skip=true` if instrumentation fails on a stale class; that failure is unrelated to this change.


manuranga:
Pairs with the ICP side: wso2/integration-control-plane#912. Neither blocks the other — the ICP flag is off by default and an MI without this ignores the tunnelled commands — but the feature needs both, so they should be released together (MI 4.7.0).

